PR #51, opened 2024-06-12
Fix URL to contributors guide

michael-the1:


jochemvandooren (approved):

README.md
```diff
@@ -42,4 +42,4 @@ Would you like to contribute to `dbt-score`? That's great news! Please follow
 
 [dbt]: https://github.com/dbt-labs/dbt-core
 [dbt-score]: https://dbt-score.picnic.tech/
-[contributors-guide]: https://dbt-score.picnic.tech/contributors_guide
+[contributors-guide]: https://dbt-score.picnic.tech/CONTRIBUTING/
```
matthieucan, line 45: Oh damn, can we configure mkdocs to use the former URL? Uppercase doesn't look good in an URL :/

michael-the1: I can't find any such configuration unfortunately

japborst: You could make a custom redirect using something like this https://github.com/mkdocs/mkdocs-redirects

michael-the1: The redirect works, but at the end of the day it's still a redirect so the URL in your address bar will still be `CONTRIBUTING/`. The only value it would bring is that the URLs in this readme look a bit nicer, but I wouldn't install a whole plugin just for that.

japborst: Ah, I had the impression it would be an url-rewrite instead of a simple redirect. Or actually
> During the mkdocs build process, this plugin will create .html files in site_dir for each of the "old" file that redirects to the "new" path

matthieucan: A simpler solution might be to rename `CONTRIBUTING.md` to `contributing.md`.

japborst: Whilst they suggest to create a file name `CONTRIBUTING` it does indeed look like lower-case is also supported
> Contributing guidelines filenames are not case sensitive

([docs](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors))

matthieucan (approved):

jochemvandooren (approved):

PR #54, opened 2024-06-12
Support Python 3.10

druzhinin-kirill:
Provide support for 3.10.

While here, fix the ci workflow to pick up Python versions properly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: ["3.11", "3.12"]
+        python: ["3.10", "3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Set up PDM
         uses: pdm-project/setup-pdm@v4
         with:
-          python-version: "3.11"
+          python-version: ${{ matrix.python }}
       - name: Install dependencies
         run: |
           pdm sync -d
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 ]
 
 dependencies = [
-    "dbt-core>=1.5, <1.8",
+    "dbt-core>=1.5",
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -15,17 +15,18 @@ classifiers = [
...
+    "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Intended Audience :: Developers",
     "Operating System :: OS Independent",
 ]
 
 dependencies = [
-    "dbt-core>=1.5",
+    "dbt-core>=1.5, <1.8",
```
druzhinin-kirill, line 26: I fixed it temporarily to see if it resolves the issue on CI.

src/dbt_score/config.py
```diff
@@ -1,13 +1,18 @@
...
 from dataclasses import dataclass, field, replace
 from pathlib import Path
 from typing import Any, Final
 
 from dbt_score.rule import RuleConfig
 
+if sys.version_info >= (3, 11):
+    import tomllib
+else:
+    import tomli as tomllib
```
matthieucan, line 14: Is it a transitive dependency? We probably should have this in `pyproject.toml` for Python < 3.11?

jochemvandooren: I think we need something like:

`tomli = { version = "...", python = "<3.11" }` in pyproject indeed

druzhinin-kirill: Yes, good point 👍

pyproject.toml
druzhinin-kirill: Okay, it did not reproduce on ci, we are safe to remove it 🆗

.github/workflows/ci.yml
```diff
@@ -18,14 +18,14 @@ jobs:
...
-        python: ["3.11", "3.12"]
+        python: ["3.10", "3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v4
       - name: Set up PDM
         uses: pdm-project/setup-pdm@v4
         with:
-          python-version: "3.11"
+          python-version: ${{ matrix.python }}
```
matthieucan, line 28: Good catch 👌

matthieucan (approved):

PR #58, opened 2024-06-16
feat: add fail_project_under and fail_any_model_under options

### LINKED ISSUE #55: Configurable acceptable project score
Hi

It would be good if we could set either 

- a minimum project score 
- minimum required severity

and have `dbt-score` return a non-zero return code so that we can stop CI pipelines. At the minute, only the inability to find a manifest file returns a non-zero return code. Duplicate rules raise an exception

e.g. for minimum required project score 

```
[tool.dbt-score]
min_required_project_score=7.5
```

or passed on command line `--min_required_project_score 7.5`

for maximum allowed severity

```
[tool.dbt-score]
min_required_severity=MEDIUM
```
or passed on command line `--min_required_severity MEDIUM`

Should it raise an exception or simply log a message and exit with non-zero return code ?

Regards
Rob

WobblyRobbly:
This pull request is for Issue #55 

Decisions
- `fail_project_under` and `fail_any_model_under` can be specified in the `pyproject.toml` or via command line
- to avoid making this a breaking change a default value of `0.0` for `fail_project_under` and `fail_any_model_under` is set
- the exit point for failing is in `cli.py` and that required the `Evaluation` instance to be returned from `lint_dbt_project`

Please let me know your thoughts of potential changes

Closes #55


tests/test_cli.py
```diff
@@ -1,5 +1,5 @@
 """Test the CLI."""
-
+import logging
```
druzhinin-kirill, line 2:
```suggestion

import logging
```

just a small style thing, not covered with our linter setup yet 😞

src/dbt_score/evaluation.py
```diff
@@ -49,6 +49,9 @@ def __init__(
         # The aggregated project score
         self.project_score: Score
 
+        # The minimum model score, starts at the maximum of 10
+        self.min_model_score: float = 10
```
druzhinin-kirill, line 53:
```suggestion
        self.min_model_score: float = 10.0
```

src/dbt_score/cli.py
```diff
@@ -102,14 +118,33 @@ def lint(
...
     if run_dbt_parse:
         dbt_parse()
 
     try:
-        lint_dbt_project(
+        evaluation = lint_dbt_project(
             manifest_path=manifest, config=config, format=format, select=select
         )
+
+        if evaluation.min_model_score < config.fail_any_model_under:
```
druzhinin-kirill, line 134: There might be some edge cases where rounding will lead to nonsensical output, like:

Given:
```
min_model_score = 8.65
fail_any_model_under = 8.69
```

```
Individual model score 8.7 is less than `fail_any_model_under` setting of 8.69
``` 

We can consider using only raw/rounded values (both in calculation and output) or making the rounding strategy a bit more advanced.

I do not think this should block this PR but it's definitely something to keep in mind 🙌

druzhinin-kirill (commented):
Thanks for your contribution!
Left some tiny comments, already looks good 🚀

tests/test_cli.py
WobblyRobbly: This has been correccted

src/dbt_score/evaluation.py
WobblyRobbly: This has been corrected

src/dbt_score/cli.py
WobblyRobbly: That makes sense. I left the rounding the same as the rounding in the actual scoring.

src/dbt_score/evaluation.py
```diff
@@ -67,6 +70,7 @@ def evaluate(self) -> None:
             self._formatter.model_evaluated(
                 model, self.results[model], self.scores[model]
             )
+            self.min_model_score = min(self.min_model_score, self.scores[model].value)
```
matthieucan, line 73: I'm not sure we should calculate the minimum here, I think there'd be more value giving a list of all models which fall below the minimum allowed instead?

src/dbt_score/config.py
```diff
@@ -67,6 +69,8 @@ def __init__(self) -> None:
         self.rules_config: dict[str, RuleConfig] = {}
         self.config_file: Path | None = None
         self.badge_config: BadgeConfig = BadgeConfig()
+        self.fail_project_under: float = 0.0
+        self.fail_any_model_under: float = 0.0
```
matthieucan, line 73: I would be in favour of a breaking change and have reasonable defaults here, e.g. `5.0`

src/dbt_score/evaluation.py
WobblyRobbly: How would something like this be ?

```🥉 customers_view (score: 7.3)
    WARN (medium) dbt_score.rules.generic.columns_have_description: Columns lack a description: c_name, c_comment.
    OK   dbt_score.rules.generic.has_description
    WARN (medium) dbt_score.rules.generic.has_owner: Model lacks an owner.
    OK   dbt_score.rules.generic.public_model_has_example_sql
    OK   dbt_score.rules.generic.sql_has_reasonable_number_of_lines

🥈 customer_nationalities (score: 9.3)
    OK   dbt_score.rules.generic.columns_have_description
    OK   dbt_score.rules.generic.has_description
    OK   dbt_score.rules.generic.has_owner
    WARN (low) dbt_score.rules.generic.public_model_has_example_sql: The model description does not include an example SQL query.
    OK   dbt_score.rules.generic.sql_has_reasonable_number_of_lines

Project score: 8.3 🥈
2024-06-17 17:55:00,214 ERROR [dbt_score.cli] Model model.view_test.customers_view scored 7.3 which is less than `fail_any_model_under` setting of 10.0
2024-06-17 17:55:00,214 ERROR [dbt_score.cli] Model model.view_test.customer_nationalities scored 9.3 which is less than `fail_any_model_under` setting of 10.0
```

matthieucan: Not sure I would use the logger, the output should match what's above to be efficiently parsed by humans.
I suggest to simplify it a bit:
```
Error: fail_any_model_under = 10.0
Model model.view_test.customers_view scored 7.3
Model model.view_test.customer_nationalities scored 9.3
```
and
```
Error: fail_project_under = 10.0
```

What do you think?

docs/configuration.md
```diff
@@ -47,6 +49,10 @@ The following options can be set in the `pyproject.toml` file:
   this setting, that the default rules are in `dbt_score.rules` and are disabled
   if not included here.
 - `disabled_rules`: A list of rules to disable.
+- `fail_project_under`: If the project score is below this value the command
+  will fail with return code 1
```
jochemvandooren, line 53:
```suggestion
  will fail with return code 1.
```

docs/configuration.md
```diff
@@ -47,6 +49,10 @@ The following options can be set in the `pyproject.toml` file:
...
+- `fail_any_model_under`: If ay model scores below this value the command will
+  fail with return code 1
```
jochemvandooren, line 55:
```suggestion
  fail with return code 1.
```

src/dbt_score/cli.py
```diff
@@ -78,6 +78,20 @@ def cli() -> None:
...
+@click.option(
+    "--fail_project_under",
+    help="Fail if the project score is under this value",
+    type=float,
+    is_flag=False,
+    default=None,
+)
+@click.option(
+    "--fail_any_model_under",
+    help="Fail if any model is under this value",
```
jochemvandooren, line 90:
```suggestion
    help="Fail if any model is under this value.",
```

src/dbt_score/cli.py
```diff
@@ -78,6 +78,20 @@ def cli() -> None:
     is_flag=True,
     default=False,
 )
+@click.option(
+    "--fail_project_under",
+    help="Fail if the project score is under this value",
```
jochemvandooren, line 83:
```suggestion
    help="Fail if the project score is under this value.",
```

src/dbt_score/cli.py
```diff
@@ -102,14 +118,35 @@ def lint(
...
+        }
+
+        if failed_models:
+            print(f"Error: fail_any_model_under = {config.fail_any_model_under}")
+            for model, value in failed_models.items():
+                print(f"Model {model} scored {round(value,1)}")
+            ctx.exit(1)
+
+        if evaluation.project_score.value < config.fail_project_under:
+            print(f"Error: fail_project_under = {evaluation.project_score.value}")
```
jochemvandooren, line 147: What about making it more in line with the failed models output?

```suggestion
            print(f"Error: fail_project_under = {config.fail_project_under}")
            print(f"Project scored {round(evaluation.project_score.value,1)}")
```

jochemvandooren (commented):
Super nice feature :rocket:, works very well. I've added some nitpicky comments, just to align with the style we have :sweat_smile: 

Looks good to me! Just missing an entry in the CHANGELOG, could you add it? :bowing_man:

druzhinin-kirill (approved):
pre-approve from my side 🚀

docs/configuration.md
WobblyRobbly: amended

docs/configuration.md
WobblyRobbly: amended

src/dbt_score/cli.py
WobblyRobbly: amended

src/dbt_score/cli.py
WobblyRobbly: amended

WobblyRobbly:
@jochemvandooren as we've added a Breaking Change i.e. will fail if the default model score isn't met (5.0) this should bump the version to 1.0.0 based on ?

```
## [1.0.0] - 2024-06-19

- Add `project_fail_under` configuration 
- Add `fail_any_model_under` configuration
- **breaking:** - default values of `5.0` for `project_fail_under` and `fail_any_model_under` will cause command to exit return code 1
```

matthieucan:
> @jochemvandooren as we've added a Breaking Change i.e. will fail if the default model score isn't met (5.0) this should bump the version to 1.0.0 based on ?
> 
> ```
> ## [1.0.0] - 2024-06-19
> 
> - Add `project_fail_under` configuration 
> - Add `fail_any_model_under` configuration
> - **breaking:** - default values of `5.0` for `project_fail_under` and `fail_any_model_under` will cause command to exit return code 1
> ```

We're following semantic versioning, meaning if the major version is `0`, a breaking change will bump the minor version.
However no need for this to happen in the changelog file for this PR, this is done when releasing :)

src/dbt_score/cli.py
```diff
@@ -102,14 +118,35 @@ def lint(
...
+            ctx.exit(1)
```
matthieucan, line 148: I think it'd make sense to have this output done by `src/dbt_score/formatters/human_readable_formatter.py`. The `Formatter` class doesn't know about the config,  but it could easily receive it upon instantiation in `__init__`. I'm happy to take care of this if you prefer - let me know.

WobblyRobbly: @matthieucan I've pushed the other PR fixes but happy for you to fix this. I'm struggling for time this week

matthieucan: @WobblyRobbly Unfortunately it seems I can't push to this PR, despite the "Maintainers are allowed to edit this pull request" option. Is there some branch protection on your side?

matthieucan: Thanks, that worked!

src/dbt_score/cli.py
```diff
@@ -102,21 +118,32 @@ def lint(
...
+
     except FileNotFoundError:
         logger.error(
             "dbt's manifest.json could not be found. If you're in a dbt project, be "
             "sure to run 'dbt parse' first, or use the option '--run-dbt-parse'."
         )
         ctx.exit(2)
 
+    if (
+        any(x.value < config.fail_any_model_under for x in evaluation.scores.values())
```
jochemvandooren, line 142: Not a big fan of using `x` personally, `score` would be more readable IMO

docs/configuration.md
```diff
@@ -47,6 +49,10 @@ The following options can be set in the `pyproject.toml` file:
   this setting, that the default rules are in `dbt_score.rules` and are disabled
   if not included here.
 - `disabled_rules`: A list of rules to disable.
+- `fail_project_under`: If the project score is below this value the command
```
jochemvandooren, line 52: Btw, I think adding the default value would be useful here! (5.0)

jochemvandooren (approved):
Very nice, approved! :raised_hands: Two comments, which I will leave up to one of you!

src/dbt_score/cli.py
matthieucan: That makes the line too long, and then spread over multiple lines I think we lose in readability

matthieucan:
@WobblyRobbly Thank you so much for your contribution! First external PR that we merged for this project 🚀 🥇 
Looking forward to more of the same!

WobblyRobbly:
Thanks for all your help on this one !!

PR #62, opened 2024-06-19
Update release workflow

druzhinin-kirill:
Use action to create GH releases automatically when a new tag is pushed.

https://github.com/ncipollo/release-action/


jochemvandooren (approved):
:exploding_head: :rocket:

matthieucan (approved):

PR #64, opened 2024-06-20
Add null check before calling project_evaluated

farmanp:
### Description
This PR adds a null check in the evaluate method of the Evaluation class before calling project_evaluated on the formatter. This ensures that project_evaluated is only called if there are models to evaluate, thus skipping the output when no model is found.

#### Changes Made
Added a null check in the evaluate method of the Evaluation class.


jochemvandooren (commented):
Looks good! 

Could you add it to the CHANGELOG? :pray:

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to
 - Add `fail_any_model_under` configuration.
 - **Breaking:** default values of `5.0` for `project_fail_under` and
   `fail_any_model_under` will cause command to exit return code 1.
+- Add null check before calling `project_evaluated` in the `evaluate` method to
```
jochemvandooren, line 19: Thanks for adding it, it needs to be under the `## [Unreleased]` section though!

farmanp: Ugh so sorry, I thought I put it under there! Thanks for your review.

jochemvandooren: Np! :grinning:

jochemvandooren (approved):
:raised_hands:

matthieucan (approved):
Wonderful, thank you for your contribution @farmanp ! 🚀

Issue #48, opened 2024-06-11
Enable/disable rules per schema


diorge:
Looking through the documentation, it seems dbt-score only allows rules to be enabled or disabled for the entire project. Would it be feasible to allow enabling/disabling of rules per folder or schema of the project? Example use case: we only enforce our silver-layer (medallion architecture) models to have an engineer as owner.

diorge:
@matthieucan Yeah that makes sense. Correct me if I'm wrong, but that also means we can specialize the generic rules in Python as well, instead of adding config to the TOML file?

```python
from dbt_score import Model, rule, RuleViolation
from dbt_score.rules.generic import has_description

@rule
def some_schema_has_description(model: Model): RuleViolation | None:
    if model.schema == 'some':
        return has_description(model)
```

matthieucan:
Yes, that's true! But for the rules provided in dbt-score (as opposed to your own rules), things should ideally apply to all (most) users.
Hence why having this configurable would be perfect, as folders/schemas are different everywhere. Hope that makes sense!

diorge:
I've been playing around with this. As much as it "works", it can skew the total score of the system by having too many passes. If a certain lint rule should only be applied to a certain models (even the generic `public_model_has_example`), you get to have a higher score by simply having non-matching models.

I'm starting to think the most elegant(?) solution is having rules return not `RuleViolation | None` but rather `RuleViolation | Bool` (or maybe `RuleViolation | SkipRule | None` to be expansible and retrocompatible).

matthieucan:
> I've been playing around with this. As much as it "works", it can skew the total score of the system by having too many passes. If a certain lint rule should only be applied to a certain models (even the generic public_model_has_example), you get to have a higher score by simply having non-matching models.

That's a good point. However, if a rule only applies to a subset of models, I don't think it's a stretch to consider other models compliant with that rule: they are not supposed to abide by a rule, so they are not violating it.

I understand it skews the results in a way, but it also keeps things simpler to understand.

Given the thresholds and rule severity can be easily configured, do you think it's an issue for you? Maybe you have some example to help me understand your use case?

diorge:
Simple example: our project has no SQL examples in any model, but none of the models are marked as public either. If I disable the generic `public_model_has_example` the project gets 6.7 total score, and by enabling it I get 7.8. One of my schemas has 20 models. If I create a rule for SQL examples for these models, using high severity, the project score goes from 6.7 to 7.6.

[...]

PR #66, opened 2024-06-27
Allow rules to be skipped

diorge:
Solves #48 (partially?)

Introduce `ModelFilter`, which allows `Rule` objects to be skipped, by filtering on a `Model`. Adding filters is done in a similar way as rules, and can be configured using the `pyproject.toml` configuration.

### DIFF
--- a/docs/create_rules.md
+++ b/docs/create_rules.md
@@ -90,3 +90,21 @@ def sql_has_reasonable_number_of_lines(model: Model, max_lines: int = 200) -> Ru
             message=f"SQL query too long: {count_lines} lines (> {max_lines})."
         )
 ```
+
+### Filtering, skipping and expanding
+
+It's possible to modify existing rules, either built-in or custom.
+A possible use case is skipping a default rule for certain models.
+Remember to also disable the original rule (see [Configuration](configuration.md)).
+
+```python
+from dbt_score import Model, rule, RuleViolation, SkipRule
+
+@rule
+def schema_x_has_description(model: Model) -> RuleViolation | SkipRule | None:
+    """Models in schema X should have a description."""
+    from dbt_score.rules.generic import has_description
+    if model.schema.lower() == 'x':
+        return SkipRule()
+    return has_description.evaluate(None, model)
+```
--- a/docs/create_rules.md
+++ b/docs/create_rules.md
@@ -39,11 +39,10 @@ from dbt_score import rule, Severity
 @rule(severity=Severity.HIGH)
 ```
 
-If a custom rule should be applied to only a subset of the models in the project,
-a special value of type `SkipRule` can be returned.
-Models that were skipped will be ignored in the final score.
-More complex uses of skipping rules can use Model Filters (see below).
-
+If a custom rule should be applied to only a subset of the models in the
+project, a special value of type `SkipRule` can be returned. Models that were
+skipped will be ignored in the final score. More complex uses of skipping rules
+can use Model Filters (see below).
 
 ```python
 from dbt_score import rule, Model, RuleViolation, SkipRule
@@ -111,14 +110,13 @@ def sql_has_reasonable_number_of_lines(model: Model, max_lines: int = 200) -> Ru
 
 ### Filtering models
 
-Custom and standard rules can be configured to have model filters.
-Filters allows setting models of the project to be ignored by a given rule.
+Custom and standard rules can be configured to have model filters. Filters
+allows setting models of the project to be ignored by a given rule.
 
-Filters are created using the same discovery mechanism and interface as custom rules,
-except they do not accept parameters.
-Similar to Python's built-in `filter` function,
-when the filter evaluation returns `True` the model should be evaluated,
-otherwise it should be skipped.
+Filters are created using the same discovery mechanism and interface as custom
+rules, except they do not accept parameters. Similar to Python's built-in
+`filter` function, when the filter evaluation returns `True` the model should be
+evaluated, otherwise it should be skipped.
 
 ```python
 from dbt_score import ModelFilter, model_filter
@@ -134,8 +132,7 @@ class SkipSchemaY(ModelFilter):
       return model.schema.lower() != 'y'
 ```
 
-Similar to setting a rule severity,
-standard rules can have filters set in the
+Similar to setting a rule severity, standard rules can have filters set in the
 [configuration file](configuration.md/#tooldbt-scorerulesrule_namespacerule_name),
 while custom rules accept the configuration file or a decorator parameter.
 
--- a/docs/create_rules.md
+++ b/docs/create_rules.md
@@ -39,6 +39,24 @@ from dbt_score import rule, Severity
 @rule(severity=Severity.HIGH)
 ```
 
+If a custom rule should be applied to only a subset of the models in the project,
+a special value of type `SkipRule` can be returned.
+Models that were skipped will be ignored in the final score.
+More complex uses of skipping rules can use Model Filters (see below).
+
+
+```python
+from dbt_score import rule, Model, RuleViolation, SkipRule
+
+@rule
+def mart_schema_has_description(model: Model) -> RuleViolation | SkipRule | None:
+    """A model in the DataMart should have a description."""
+    if model.schema != 'mart':
+        return SkipRule()
+    if not model.description:
+        return RuleViolation(message="Model lacks a description.")
+```
+
 ### Using the `Rule` class
 
 For more advanced use cases, a rule can be created by inheriting from the `Rule`
@@ -91,30 +109,16 @@ def sql_has_reasonable_number_of_lines(model: Model, max_lines: int = 200) -> Ru
         )
 ```
 
-### Filtering
+### Filtering models
 
-Custom and standard rules can be configured to have model filters. Filters
-allows setting models of the project to be ignored by a given rule.
+Custom and standard rules can be configured to have model filters.
+Filters allows setting models of the project to be ignored by a given rule.
 
-Custom rules can skip models "manually":
-
-```python
-from dbt_score import rule, RuleViolation, SkipRule
-
-@rule
-def models_in_x_follow_naming_standard(model: Model) -> RuleViolation | SkipRule | None:
-    if model.schema.lower() != 'x':
-        return SkipRule()
-    if some_regex_fails(model.name):
-        return RuleViolation("Invalid model name.")
-```
-
-Filters can also be set as a first-class citizen. Doing so allow re-use and
-composability, as well as being used by generic rules. First, filters should be
-created using the same discovery mechanism and interface as custom rules, except
-they do not accept parameters. Similar to Python's built-in `filter` function,
-when the filter returns `True` the model should be evaluated, otherwise it
-should be skipped.
+Filters are created using the same discovery mechanism and interface as custom rules,
+except they do not accept parameters.
+Similar to Python's built-in `filter` function,
+when the filter evaluation returns `True` the model should be evaluated,
+otherwise it should be skipped.
 
 ```python
 from dbt_score import ModelFilter, model_filter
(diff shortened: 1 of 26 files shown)


docs/create_rules.md
```diff
@@ -90,3 +90,21 @@ def sql_has_reasonable_number_of_lines(model: Model, max_lines: int = 200) -> Ru
             message=f"SQL query too long: {count_lines} lines (> {max_lines})."
         )
 ```
+
+### Filtering, skipping and expanding
+
+It's possible to modify existing rules, either built-in or custom.
+A possible use case is skipping a default rule for certain models.
+Remember to also disable the original rule (see [Configuration](configuration.md)).
```
matthieucan, line 98: I'm quite on the fence with this approach. It requires a 2-step process to be followed, disabling a rule and creating a wrapper for it, which is quite involved and complex.

I understand you explored the alternative approach of evaluating an expression stored in the config file. Was there any blocker with it?

matthieucan (commented):
Thank you for your contribution! 🙏 

Left a comment for discussion on the overall design of the feature.

docs/create_rules.md
diorge: The "blocker" is my unwillingness to use `eval()` in the code, really. Safely parsing an expression like `skip_if = "(model.schema.lower() == 'x') or (model.schema.lower() == 'y' and model.access == 'public')` is quite a large task.

If we just `eval()` it is of course not hard to implement. Which to be fair, is not really that much unsafer than using user-defined Python code. But at least from an user perspective, it is clearly Python code and not some configuration file.

jochemvandooren: Hey @diorge, first of all thanks for opening the PR :raised_hands: I'd like to pitch in a little bit on the design of things here. I agree with both of you: using `eval` is not too nice and I would love users to write python code in a python file (instead of a config file). :grin: However, this solution is also not perfect IMO, because it requires users to recreate a rule in order to configure skipping. It would be great if we could just add skip configuration to existing rules, without having to recreate them and at the same time write the skip logic in Python.

I think I have an idea that would solve all concerns, please think along! What if we do the following: we implement something similar to `@rule`; let's call it `@skip` for now. `@skip` will be discovered in a similar way as `@rule` and can be defined by users themselves like this:

```python
from dbt_score import Model, skip

@skip
def temp_schema(model: Model) -> bool:
    """Skip models in the temp schema"""
    if model.schema == "temp":
        return True
    return False
```

By doing this, the skip logic is entirely written in Python, and it follows the approach we take for rules! So it's quite easy IMO.

Then, to configure the skips, they can be configured on rule level and re-used! We could configure like this:

```toml
[tool.dbt-score.rules."generic.has_description"]
skips = ["temp_schema"]
```

I think this would solve the biggest concerns we have with the current two proposal:
- Not writing Python in python files
- No use of `eval()`
- Not having to disable + recreate rules to add skip logic
- Not being able to re-use skip-logic in all rules

This will be harder to implement, but it's the UX that counts if you ask me! Please let me hear your opinions! Happy to collaborate on this by the way

jochemvandooren:
> I'm currently working on updating this with the Skip feature as well. Just externalizing here that having parametrized skips work well enough as long as they are used programatically and not through the configfile interface. I thought about having generic skips such as "skip_schema(schema)" and "keep_schema(schema)", but that would not work nicely with the configfile.
> 
> On that note, Skip is not a great name because inverting it is also a valid use case. Any other name recommendation? RuleFilter?

Great to hear @diorge! Yes I think it will be hard to configure these properly with the config files, let's keep it simple (for now)!

Could you maybe elaborate on "inverting it"? That's not entirely clear to me. Anyway: `RuleFilter` is also a good name to be honest, or maybe `ModelFilter`, as it's actually filtering out some models

diorge:
> Could you maybe elaborate on "inverting it"?

Both "apply only on schema X" and "apply everywhere except schema X" are valid filters. One is skipping X, and the other is the invert operation.

Anyway, yeah, `ModelFilter` is a much better name for the feature, thanks for the input.

src/dbt_score/model_filter.py
```diff
@@ -0,0 +1,115 @@
...
+        """Initialize the filter."""
+        pass
+
+    def __init_subclass__(cls, **kwargs) -> None:  # type: ignore
+        """Initializes the subclass."""
+        super().__init_subclass__(**kwargs)
+        if not hasattr(cls, "description"):
+            raise AttributeError("Subclass must define class attribute `description`.")
+
+    def evaluate(self, model: Model) -> bool:
```
jochemvandooren, line 25: I think it would make sense to rename `evaluate` for `ModelFilter`. Maybe just `filter`?

src/dbt_score/model_filter.py
```diff
@@ -0,0 +1,115 @@
...
+        # Get description parameter, otherwise use the docstring
+        filter_description = description or (
+            func.__doc__.split("\n")[0] if func.__doc__ else None
+        )
+
+        def wrapped_func(self: ModelFilter, *args: Any, **kwargs: Any) -> bool:
+            """Wrap func to add `self`."""
+            return func(*args, **kwargs)
+
+        # Create the filter class inheriting from Skip
```
jochemvandooren, line 93: Inheriting from Skip? :thinking:

src/dbt_score/rule.py
```diff
@@ -73,7 +82,7 @@ def process_config(self, rule_config: RuleConfig) -> None:
 
         # Overwrite default rule configuration
         for k, v in rule_config.config.items():
-            if k in self.default_config:
+            if k in self.default_config or k == "model_filter_names":
```
jochemvandooren, line 85: This `model_filter_names` can be implemented in the `RuleConfig` class. It should be treated similarly to `Severity`

jochemvandooren: Then it can be added to `Rule` later on in `process_config`, so it becomes a Rule property

docs/create_rules.md
```diff
@@ -90,3 +90,60 @@ def sql_has_reasonable_number_of_lines(model: Model, max_lines: int = 200) -> Ru
             message=f"SQL query too long: {count_lines} lines (> {max_lines})."
         )
 ```
+
+### Filtering
```
jochemvandooren, line 94: There's many ways to filter! I would say we try to nudge people into using one way. IMO that would be using the `pyproject.toml` file. So people would create a `@model_filter` and apply it to a rule in `pypoject.toml`. Or do you think using `SkipRule()` and `@rule(model_filters={only_schema_x()})` is more straightforward?

My main concern is that, depending on if you own the rule or not, the way of filtering changes

jochemvandooren (commented):
Amazing work :rocket: It works super well and code looks good. I have left a couple of suggestions and questions. Will take another look later this week!

docs/create_rules.md
diorge: There are three ways of filtering: two are shared with Severity (`pyproject.toml` and inside `@rule`). The other is the feature of a custom rule returning `SkipRule()`.

If we are going to change it, I'd say dropping the latter is better, so we keep parity with Severity. However I personally like this feature, it may feel cumbersome to create a `ModelFilter` that is only going to be used once by a custom rule.

Maybe I just wrote the documentation too dense, and it can be seen as a separate thing. Or maybe it is in fact too much and we should drop it. I'll let you decide on this.

src/dbt_score/model_filter.py
diorge: I don't really like `evaluate` either, just kept consistent with `Rule`. Not opposed to changing it at all. However `filter` is not a great name either: `filter(Callable[[...], bool], Iterable[T]) -> Iterable[T]`, and `evaluate(...) -> bool` is just the first parameter of filter, not the higher-order filter function.

That is to say, the name filter for this function conflicts with the semantics of Python's built-in `filter`.

jochemvandooren: Right `filter` is not a good idea either! :sweat_smile: Maybe @matthieucan has a good idea?

docs/create_rules.md
jochemvandooren: If you like the feature, I think we should not drop it! I like the idea of documenting it as a separate thing :+1:

diorge:
My bad on not having installed the pre-commit hooks, the linter should be passing now

druzhinin-kirill:
Hey @diorge !

Thanks for this work, a really great feature 💪 

I see now there are two mechanisms to disable evaluation for a model-rule pair. Is there any practical difference between them so we should keep both? For me, it feels like `ModelFilter` provides all the functionality to achieve skipping. It also keep rule logic abstracted and seems more natural (verify if the rule should be applied before applying). Do you think the `SkipRule()` mechanism needs to be kept?

diorge:
@druzhinin-kirill this has been brought up in a conversation marked as resolved. `SkipRule` could be removed, but makes some use cases easier to deal with. I'd argue for most use cases, having a simple `return SkipRule()` is simpler than creating a ModelFilter, and filters won't be reusable by much.

I'm not too strong for keeping it, but I do think it's a nice addition that does not bloat the lib too much.

druzhinin-kirill:
> @druzhinin-kirill this has been brought up in a conversation marked as resolved. `SkipRule` could be removed, but makes some use cases easier to deal with. I'd argue for most use cases, having a simple `return SkipRule()` is simpler than creating a ModelFilter, and filters won't be reusable by much.
> 
> I'm not too strong for keeping it, but I do think it's a nice addition that does not bloat the lib too much.

I see, got your points 👍 
However, IMO the rule logic should be separated from the filter. Rules should left untouched if filters change. `ModelFilter` seems pretty powerful and clear from the design perspective, so following
```
There should be one-- and preferably only one --obvious way to do it.
```
I would vote to keep filters only. They might have a little overhead compared to `SkipRule` in specific cases indeed, but still straightforward enough to understand and implement 💡 

WDYT @jochemvandooren @matthieucan ?

src/dbt_score/model_filter.py
druzhinin-kirill: No strong opinion, but can suggest `apply` for consideration.

matthieucan:
> > @druzhinin-kirill this has been brought up in a conversation marked as resolved. `SkipRule` could be removed, but makes some use cases easier to deal with. I'd argue for most use cases, having a simple `return SkipRule()` is simpler than creating a ModelFilter, and filters won't be reusable by much.
> > I'm not too strong for keeping it, but I do think it's a nice addition that does not bloat the lib too much.
> 
> I see, got your points 👍 However, IMO the rule logic should be separated from the filter. Rules should left untouched if filters change. `ModelFilter` seems pretty powerful and clear from the design perspective, so following
> 
> ```
> There should be one-- and preferably only one --obvious way to do it.
> ```
> 
> I would vote to keep filters only. They might have a little overhead compared to `SkipRule` in specific cases indeed, but still straightforward enough to understand and implement 💡
> 
> WDYT @jochemvandooren @matthieucan ?

I agree with that sentiment. Separating rules (pure logic) and configuration (to whom does that logic apply, with which parameters) will keep things simple, and avoid some edge-cases that do not warrant the extra complexity IMO.

src/dbt_score/model_filter.py
matthieucan: I think `evaluate` is pretty good, as it's descriptive of what's happening. This function answers a yes/no question about a given model, which is fundamentally an evaluation?
`apply` works as well!

docs/create_rules.md
```diff
@@ -90,3 +107,42 @@ def sql_has_reasonable_number_of_lines(model: Model, max_lines: int = 200) -> Ru
...
+
+@model_filter
+def only_schema_x(model: Model) -> bool:
+    """Only applies a rule to schema X."""
+    return model.schema.lower() == 'x'
+
+class SkipSchemaY(ModelFilter):
+    description = "Applies a rule to every schema but Y."
+    def evaluate(self, model: Model) -> bool:
+      return model.schema.lower() != 'y'
```
matthieucan, line 132: Do we know use-cases where the class version (as opposed to the decorator version) would be necessary?

jochemvandooren: Not yet, but no harm in having it as it does not really change implementation and it follows the same design of `@rule`.

jochemvandooren:
@diorge I made some changes to the PR! Mainly got rid of the `SkipRule`, everything else is the same. @druzhinin-kirill @matthieucan Please have another look so we can get this one merged :raised_hands:

src/dbt_score/evaluation.py
```diff
@@ -57,11 +57,11 @@ def evaluate(self) -> None:
             self.results[model] = {}
             for rule in rules:
                 try:
-                    result: RuleViolation | None = rule.evaluate(model, **rule.config)
+                    if rule.should_evaluate(model):  #  Consider model filter(s).
```
druzhinin-kirill, line 60: IIUC, if a filter skips the model, it's not now visible to the user. In the case of complex filters, it might be important to flag skipping so the filter behaves expectedly 

It might be interesting to either:
* emit a warning if the model skipped (maybe too much noise)
* Keep track of skipped models in the skipped list attr. In the end, `Formatter` can use this to print at least the number of skipped models.

My proposal is to add `self.skips` and track Model, Filter

I am fine to outline it as a new feature request.

druzhinin-kirill (approved):
Great work @diorge! Thanks a lot for bringing life 🚀

src/dbt_score/evaluation.py
jochemvandooren: Yes let's keep it for another PR, then we can already release this one :rocket:

jochemvandooren:
@diorge Thanks a lot for your contribution! :raised_hands:

PR #68, opened 2024-08-07
Add JSON formatter for machine-readable output

matthieucan:
The ability to save and use results from dbt-score in other tooling opens new possibilities.
To that end, add a new formatter, which uses `stdout` to print a JSON document displaying `dbt-score`'s results.

Example:
```shell
$ dbt_score lint -f json
{
  "models": {
    "model1": {
      "score": 5.0,
      "badge": "🥈",
      "pass": true,
      "results": {
        "dbt_score.rules.generic.columns_have_description": {
          "result": "OK",
          "severity": "medium",
          "message": null
        },
        "dbt_score.rules.generic.has_description": {
          "result": "WARN",
          "severity": "medium",
          "message": "Model lacks a description."
        }
      }
    }
  },
  "project": {
    "score": 5.0,
    "badge": "🥈"
    "pass": true
  }
}
```

### DIFF
--- a/src/dbt_score/formatters/json_formatter.py
+++ b/src/dbt_score/formatters/json_formatter.py
@@ -71,7 +71,7 @@ class JSONFormatter(Formatter):
             if result is None:
                 self._model_results[model.name]["results"][rule.source()] = {
                     "result": "OK",
-                    "severity": None,
+                    "severity": rule.severity.name.lower(),
                     "message": None,
                 }
             elif isinstance(result, RuleViolation):
@@ -83,6 +83,7 @@ class JSONFormatter(Formatter):
             else:
                 self._model_results[model.name]["results"][rule.source()] = {
                     "result": "ERR",
+                    "severity": rule.severity.name.lower(),
                     "message": str(result),
                 }
 
--- a/src/dbt_score/formatters/json_formatter.py
+++ b/src/dbt_score/formatters/json_formatter.py
@@ -5,7 +5,7 @@ Shape of the JSON output:
 {
     "models": {
         "model_foo": {
-            "score": 0.5,
+            "score": 5.0,
             "badge": "🥈",
             "results": {
                 "rule1": {
@@ -32,7 +32,7 @@ Shape of the JSON output:
         }
     },
     "project": {
-        "score": 0.25,
+        "score": 2.5,
         "badge": "🥉"
     }
 }
--- a/src/dbt_score/formatters/json_formatter.py
+++ b/src/dbt_score/formatters/json_formatter.py
@@ -7,6 +7,7 @@ Shape of the JSON output:
         "model_foo": {
             "score": 5.0,
             "badge": "🥈",
+            "pass": true,
             "results": {
                 "rule1": {
                     "result": "OK",
@@ -23,6 +24,7 @@ Shape of the JSON output:
         "model_bar": {
             "score": 0.0,
             "badge": "🥉",
+            "pass": false,
             "results": {
                 "rule1": {
                     "result": "ERR",
@@ -33,7 +35,8 @@ Shape of the JSON output:
     },
     "project": {
         "score": 2.5,
-        "badge": "🥉"
+        "badge": "🥉",
+        "pass": false
     }
 }
 """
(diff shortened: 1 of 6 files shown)


src/dbt_score/formatters/json_formatter.py
```diff
@@ -0,0 +1,100 @@
+"""JSON formatter.
+
+Shape of the JSON output:
+
+{
+    "models": {
+        "model_foo": {
+            "score": 0.5,
```
jochemvandooren, line 8:
```suggestion
            "score": 5.0,
```

src/dbt_score/formatters/json_formatter.py
```diff
@@ -0,0 +1,100 @@
...
+            "results": {
+                "rule1": {
+                    "result": "ERR",
+                    "message": "Exception message"
+                }
+            }
+        }
+    },
+    "project": {
+        "score": 0.25,
```
jochemvandooren, line 35:
```suggestion
        "score": 2.5,
```

src/dbt_score/formatters/json_formatter.py
```diff
@@ -0,0 +1,100 @@
...
+        """Callback when a model has been evaluated."""
+        self._model_results[model.name] = {
+            "score": score.value,
+            "badge": score.badge,
+            "results": {},
+        }
+        for rule, result in results.items():
+            if result is None:
+                self._model_results[model.name]["results"][rule.source()] = {
+                    "result": "OK",
```
jochemvandooren, line 73: Maybe we can make this an Enum or something, so it's clear what are the possible results codes?

src/dbt_score/formatters/json_formatter.py
```diff
@@ -0,0 +1,100 @@
...
+                    "severity": rule.severity.name.lower(),
+                    "message": None,
+                }
+            elif isinstance(result, RuleViolation):
+                self._model_results[model.name]["results"][rule.source()] = {
+                    "result": "WARN",
+                    "severity": rule.severity.name.lower(),
```
jochemvandooren, line 80: We could set `severity = rule.severity.name.lower()` above and re-use it

src/dbt_score/formatters/json_formatter.py
```diff
@@ -0,0 +1,100 @@
...
+        """Instantiate formatter."""
+        super().__init__(*args, **kwargs)
+        self._model_results: dict[str, dict[str, Any]] = {}
+        self._project_results: dict[str, Any]
+
+    def model_evaluated(
+        self, model: Model, results: ModelResultsType, score: Score
+    ) -> None:
+        """Callback when a model has been evaluated."""
+        self._model_results[model.name] = {
```
jochemvandooren, line 65: I think it would be valuable to add something like `above_threshold` on a `model` level. I think it's a property that is actionable; a `model` can have a certain badge + score, but still considered 'not good enough' based on the thresholds configurations.

jochemvandooren (commented):
Looks very good! :rocket:

src/dbt_score/formatters/json_formatter.py
matthieucan: Good point! I added a `pass` boolean property to both models and project, PTAL!

tests/formatters/test_json_formatter.py
```diff
@@ -0,0 +1,64 @@
...
+        },
+        "tests.conftest.rule_severity_medium": {
+          "result": "ERR",
+          "severity": "medium",
+          "message": "Oh noes"
+        },
+        "tests.conftest.rule_severity_critical": {
+          "result": "WARN",
+          "severity": "critical",
+          "message": "Error"
```
jochemvandooren, line 52: Interesting `result` considering the `rules.result` :grin: Not a problem though!

jochemvandooren (approved):
Awesome! :rocket: I'll leave the `Enum` comment up to you :+1:

src/dbt_score/formatters/json_formatter.py
matthieucan: That's a good idea, but given it's also used in the `HumanReadableFormatter`, this would require a bit of refactoring to have it proper. I suggest to keep it simple for now :)

PR #73, opened 2024-08-21
Improve error handling in CLI

matthieucan:
Improve error handling:
- If `dbt parse` fails to run, a nice(r) error message is displayed
- In case of any error, the return code is 2
- Logging is written to stderr, instead of stdout

### DIFF
--- a/src/dbt_score/cli.py
+++ b/src/dbt_score/cli.py
@@ -1,6 +1,7 @@
 """CLI interface."""
 
 import logging
+import traceback
 from pathlib import Path
 from typing import Final, Literal
 
@@ -9,7 +10,7 @@ from click.core import ParameterSource
 from dbt.cli.options import MultiOption
 
 from dbt_score.config import Config
-from dbt_score.dbt_utils import dbt_parse, get_default_manifest_path
+from dbt_score.dbt_utils import DbtParseException, dbt_parse, get_default_manifest_path
 from dbt_score.lint import lint_dbt_project
 from dbt_score.rule_catalog import display_catalog
 
@@ -123,10 +124,9 @@ def lint(
     if fail_any_model_under:
         config.overload({"fail_any_model_under": fail_any_model_under})
 
-    if run_dbt_parse:
-        dbt_parse()
-
     try:
+        if run_dbt_parse:
+            dbt_parse()
         evaluation = lint_dbt_project(
             manifest_path=manifest, config=config, format=format, select=select
         )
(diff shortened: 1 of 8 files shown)


src/dbt_score/cli.py
```diff
@@ -138,6 +138,14 @@ def lint(
         )
         ctx.exit(2)
 
+    except DbtParseException:
+        logger.error("dbt failed to parse project. Run `dbt parse` to investigate.")
+        ctx.exit(2)
+
+    except Exception:
+        logger.error(traceback.format_exc())
+        ctx.exit(2)
```
druzhinin-kirill, line 147: I think it might not be 100% correct:

> Exit code 2 signifies invalid usage of some shell built-in command.

While here, the exception might be propagated from `lint_dbt_project` and relate to any internal bug. 
I think we can avoid broad exception handling and allow it propagate as is.

matthieucan: I want to distinguish those 3 situations:
- Linter ran, and project is ok (warnings below threshold)
- Linter ran, and project is not ok (warnings above threshold)
- Unexpected error

The first one should be 0. I'm open to any exit code for the second and third situations, but they should be different. This approach is quite common, see e.g. black: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#check

Would you suggest something else than 1 and 2?

druzhinin-kirill: I believe broad exception handling should result in exit code 1. Exit code 2 is a bit misleading IMO as it identifies an invalid command call (while here a command is valid).
We build on top of `click` that follows this and has 2 exit code for [UsageError](https://github.com/pallets/click/blob/123dd717439d8620d8d6be5574d2c9f007952326/click/exceptions.py#L43). So 2 is sort of "reserved" for a specific meaning.

However, I checked other tools and it's not a strict standard and some tools have their own exit code interpretation. Also, checked a good explanation for [Ruff](https://docs.astral.sh/ruff/linter/#exit-codes) exit codes. 
So indeed here, they use 2 for any kind of errors (like you implemented initially) 👌 

TL/DR everything works if we document it 😄

druzhinin-kirill (approved):

src/dbt_score/cli.py
matthieucan: Indeed, POSIX only specifies 0 for success and >0 for failure.
On top of this, it's only conventions and/or specific decisions by programs.
Glad to hear Ruff follows the same, I'm pretty sure I've seen this in other linters too, but we can adapt this convention as well, and revise later if we find it causes issues.

I'll document it!

matthieucan: Done, PTAL :)

jochemvandooren:
Nice :rocket:

PR #74, opened 2024-08-22
Auto-round scores down to align scores & medals

### LINKED ISSUE #61: Bug: Silver medal with 10.0 project score
It appears one can get a `10.0` _project_ score, but still a silver medal
image

My suspicion is that the score is actually rounded (e.g. 9.95).

japborst:
Fixes #61.

The challenge is that a 9.99 score leads to a silver medal, even though 10.0 is printed as in `HumanReadableFormatter` we're rounding to a single decimal.

There are two approaches to this problem
- Fix in the human formatter
- Round differently everywhere

Separately, you could argue this is only a problem for scores near thresholds and not in other instances (e.g. 3.35 might be fine to round to 3.4, but 9.99 should be displayed as 9.9).

To keep things simple, I've added a property on the `Score` model and used the rounded-down version in the human-readable formatter.

### DIFF
--- a/src/dbt_score/formatters/human_readable_formatter.py
+++ b/src/dbt_score/formatters/human_readable_formatter.py
@@ -33,7 +33,7 @@ class HumanReadableFormatter(Formatter):
         """Callback when a model has been evaluated."""
         if score.value < self._config.fail_any_model_under:
             self._failed_models.append((model, score))
-        print(f"{score.badge} {self.bold(model.name)} (score: {score.human_value!s})")
+        print(f"{score.badge} {self.bold(model.name)} (score: {score.rounded_value!s})")
         for rule, result in results.items():
             if result is None:
                 print(f"{self.indent}{self.label_ok} {rule.source()}")
@@ -48,7 +48,7 @@ class HumanReadableFormatter(Formatter):
 
     def project_evaluated(self, score: Score) -> None:
         """Callback when a project has been evaluated."""
-        print(f"Project score: {self.bold(str(score.human_value))} {score.badge}")
+        print(f"Project score: {self.bold(str(score.rounded_value))} {score.badge}")
 
         if len(self._failed_models) > 0:
             print()
--- a/src/dbt_score/formatters/human_readable_formatter.py
+++ b/src/dbt_score/formatters/human_readable_formatter.py
@@ -33,7 +33,7 @@ class HumanReadableFormatter(Formatter):
         """Callback when a model has been evaluated."""
         if score.value < self._config.fail_any_model_under:
             self._failed_models.append((model, score))
-        print(f"{score.badge} {self.bold(model.name)} (score: {score.value!s})")
+        print(f"{score.badge} {self.bold(model.name)} (score: {score.human_value!s})")
         for rule, result in results.items():
             if result is None:
                 print(f"{self.indent}{self.label_ok} {rule.source()}")
@@ -48,7 +48,7 @@ class HumanReadableFormatter(Formatter):
 
     def project_evaluated(self, score: Score) -> None:
         """Callback when a project has been evaluated."""
-        print(f"Project score: {self.bold(str(score.value))} {score.badge}")
+        print(f"Project score: {self.bold(str(score.human_value))} {score.badge}")
 
         if len(self._failed_models) > 0:
             print()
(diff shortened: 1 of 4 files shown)


src/dbt_score/scoring.py
```diff
@@ -19,6 +20,10 @@ class Score:
     value: float
     badge: str
 
+    def __post_init__(self) -> None:
+        """Auto-round score down to 1 decimal place."""
+        self.value = math.floor(self.value * 10) / 10
```
matthieucan, line 25: Instead of overriding the provided score upon instantiation, I would suggest creating a separate `@property`, i.e. a derived value. This would allow:
- Keeping the original value
- Keeping things correct when mutated (e.g. `score = Score(9.99, 'foo'); score.value = 5`)

WDYT?

japborst: Yep, that's definitely also something that can work.

Would you then suggest we only use the property (e.g. `rounded_value`) in the plain (human readable) output?

matthieucan: Yes, and maybe keep the non-rounded in the JSON output? Although we could also use the rounded one in JSON, for consistency purposes 🤔

japborst: Let's keep the non-rounded in the machine-readable outputs 👍

japborst: I've updated the PR to reflect this.

src/dbt_score/formatters/human_readable_formatter.py
```diff
@@ -51,16 +48,16 @@ def model_evaluated(
...
 
         if len(self._failed_models) > 0:
             print()
             print(
                 f"Error: model score too low, fail_any_model_under = "
                 f"{self._config.fail_any_model_under}"
             )
-            for model, score in self._failed_models:
-                print(f"Model {model.name} scored {round(score.value, 1)}")
+            for model, model_score in self._failed_models:
```
japborst, line 59: We were redefining `score` here. So whilst here, I've updated this.

src/dbt_score/scoring.py
```diff
@@ -19,6 +20,11 @@ class Score:
     value: float
     badge: str
 
+    @property
+    def human_value(self) -> float:
```
japborst, line 24: Clarified that this is only for human-readable purpose (instead of naming it e.g. `rounded_value`)

matthieucan: IMO it's clearer to use names that relate to what it is, rather than the intended use. That would make it more descriptive, and suitable for other usages. No strong opinion though

tests/formatters/test_human_readable_formatter.py
```diff
@@ -29,10 +28,10 @@ def test_human_readable_formatter_model(
...
     assert (
         stdout
-        == """🥇 \x1B[1mmodel1\x1B[0m (score: 10.0)
-    \x1B[1;32mOK  \x1B[0m tests.conftest.rule_severity_low
-    \x1B[1;31mERR \x1B[0m tests.conftest.rule_severity_medium: Oh noes
-    \x1B[1;33mWARN\x1B[0m (critical) tests.conftest.rule_severity_critical: Error
+        == """🥇 \x1b[1mmodel1\x1b[0m (score: 10.0)
+    \x1b[1;32mOK  \x1b[0m tests.conftest.rule_severity_low
+    \x1b[1;31mERR \x1b[0m tests.conftest.rule_severity_medium: Oh noes
+    \x1b[1;33mWARN\x1b[0m (critical) tests.conftest.rule_severity_critical: Error
```
matthieucan, line 34: Interesting, I know hex is not case sensitive, but curious how this changed 🤔

src/dbt_score/scoring.py
japborst: Would you have a more suitable suggestion for a name?

matthieucan: I think `rounded_value` as you mentioned above, or `floored_value`, would work well

tests/formatters/test_human_readable_formatter.py
japborst: Appears it's my local `ruff` kicking in action.

Even though ruff is configured for this project, it's on 0.2.2 and locally have 0.6.1 running.

Will revert this change.

matthieucan (approved):
🎉

jochemvandooren:
Thanks for fixing it @japborst ! :raised_hands:

PR #77, opened 2024-10-02
Only output failing models and violated rules per default in HumanReadableFormatter

### LINKED ISSUE #71: Add lint option to only output models with score under some threshold
We are mostly interested in knowing which models are not up to standard, and that can be hard to glimpse when the project has many models and most of them already have a gold badge.

It would be very useful to have an option for the `lint` command that will cause it to only output models whose score is less than the first threshold (or perhaps more flexible, a custom threshold given as argument to the option.)

Related to this, it would also be useful to output only the rules that were not satisfied, but I'm unsure if that should be a separate option or a direct result of the one suggested.

thomend:
Fixes #71 

It changes the default behavior of the HumanReadableFormatter:
When running `dbt-score lint` it will now only show failing models and/or rules which were violated.
To have all output printed (exactly the way it was before this PR) the user has now to pass the `--show_all` flag to `dbt-score lint`. 
It also considers the `fail_any_model_under` option. 

Furthermore, I had to adjust some of tests since the default behavior of the HumanReadableFormatter is now changed - let me know if I would need to expand this. 

The docs also reflect this change by slightly tweaking the first example in `index.md`

In this current form, both rule violations and failing models are filtered through the same argument. This is based on @matthieucan. 's comment on https://github.com/PicnicSupermarket/dbt-score/issues/71#issuecomment-2383106560


src/dbt_score/cli.py
```diff
@@ -93,8 +93,17 @@ def cli() -> None:
     is_flag=False,
     default=None,
 )
+@click.option(
+    "--show_all",
```
matthieucan, line 97:
```suggestion
    "--show-all",
```
Let's use a dash for the CLI

docs/index.md
```diff
@@ -21,7 +21,8 @@ Score: 10.0 🥇
 
 In this example, the model `customers` scores the maximum value of `10.0` as it
 passes all the rules. It also is awarded a golden medal because of the perfect
-score.
+score. By default a passing model without rule violations will only be shown if
+we pass the `---show_all` flag.
```
jochemvandooren, line 25:
```suggestion
score. By default a passing model with or without rule violations will not be shown, 
unless we pass the `--show-all` flag.
```

src/dbt_score/cli.py
```diff
@@ -93,8 +93,17 @@ def cli() -> None:
     is_flag=False,
     default=None,
 )
+@click.option(
+    "--show-all",
+    help="If set to True,show all models and all rules in output"
+    "when using `plain` as `--format`."
+    "Default behavior is to only show failing models and violated rules",
```
jochemvandooren, line 100:
```suggestion
    help="If set to True, show all models and all rules in output "
    "when using `plain` as `--format`. "
    "Default behavior is to only show failing models and violated rules.",
```

jochemvandooren: To fix this:

image

src/dbt_score/formatters/human_readable_formatter.py
```diff
@@ -33,18 +33,30 @@ def model_evaluated(
...
-                    f"{self.indent}{self.label_warning} "
-                    f"({rule.severity.name.lower()}) {rule.source()}: {result.message}"
-                )
-            else:
-                print(f"{self.indent}{self.label_error} {rule.source()}: {result!s}")
-        print()
+        if (
+            score.value < self._config.fail_any_model_under
+            or any(isinstance(result, RuleViolation) for result in results.values())
+            or self._config.show_all
```
jochemvandooren, line 39: I think this is not 100% correct :thinking: If I understand correctly we want to print the following output if:

`score.value < self._config.fail_any_model_under`. Then, only show the failing rules. Now it will also show the failing rules of models that did not fail

tests/formatters/test_human_readable_formatter.py
```diff
@@ -44,7 +75,7 @@ def test_human_readable_formatter_project(capsys, default_config, manifest_loade
     )
     formatter.project_evaluated(Score(10.0, "🥇"))
     stdout = capsys.readouterr().out
-    assert stdout == "Project score: \x1B[1m10.0\x1B[0m 🥇\n"
+    assert stdout == "Project score: \x1b[1m10.0\x1b[0m 🥇\n"
```
jochemvandooren, line 78: Why did all the B's turn into b? :grin:

jochemvandooren (commented):
Thank you for your contribution! :raised_hands: Looks in good shape already, just some minor comments!

src/dbt_score/cli.py
thomend: Implemented!

src/dbt_score/formatters/human_readable_formatter.py
jochemvandooren: Ok I got it! :+1:  I think for the scenario you describe it is indeed useful to be able to show all the failing rules and we can definitely leave that as the default. I do think that we should also have the option to show only failing models, with their failing rules.

Maybe we should have two flags: `--show-all-rules` and `--show-all-models` so the user is able to further specify the output. @matthieucan curious to hear your opinion as well!

So then the user is able to:
1. show failing models, with all rules
2. show failing models, with failing rules
3. show all models, with all rules
4. show all models, with failing rules

matthieucan: Good to align on those expectations indeed.
Maybe the `--show` parameter could take an argument, e.g.
- `--show all` - show all models, all rules
- `--show failing-models` - show failing rules of failing models
- `--show failing-rules` - show failing rules of all models (the default?)

I'm not sure if the first scenario mentioned (show failing models, with all rules) is useful, considering the option to use `--select` in combination. For example `--select my_model --show all` might be more actionable. But let me know what you think :)

jochemvandooren: You are right that option 1 (failing models with all rules) is probably not very useful. I think the direction of `--show something` would be a nice one. It's indeed simpler than providing two flags. And agreed that `--show failing-rules` should be the default!

thomend: Sounds good and happy to adjust it in the next few days - thanks for all the input.

thomend: In terms of wording: The codebase often refers to "violated" rules, e.g. `RuleViolation` and not failing rules. What do you think of:
`--show all`
`--show failing-models`
`--show violated-rules`
?

matthieucan: I think it's simpler to remember for users if the same term is used for the CLI options, as it's an abstraction over the code base

src/dbt_score/cli.py
```diff
@@ -93,8 +93,17 @@ def cli() -> None:
...
 )
+@click.option(
+    "--show-all",
+    help="If set to True, show all models and all rules in output "
+    "when using `plain` as `--format`. "
+    "Default behavior is to only show failing models and violated rules.",
+    type=bool,
+    is_flag=True,
+    default=False,
+)
```
matthieucan, line 104: Let's not make it a flag, but rather a parameter with a value? i.e. `--show all`, which can later be expanded into different options such as `--show failing-models` and the like

src/dbt_score/formatters/human_readable_formatter.py
thomend: Hi @jochemvandooren  sorry not picking it up earlier! I initially planned to work on it much sooner again. I am picking it up today evening again and will come back to you asap then - hope that works for you.

thomend: I just added the different options as discussed above. Will now work on the merge conflicts :)

jochemvandooren: No problem @thomend ! Was afraid you had forgotten :cry: I will review tomorrow!

src/dbt_score/cli.py
```diff
@@ -94,8 +94,20 @@ def cli() -> None:
     is_flag=False,
     default=None,
 )
+@click.option(
+    "--show",
+    help="Type of output which should be shown"
+    "when using `plain` as `--format`. "
+    "`all` shows all models and all rules"
+    "`failing-models` shows failing rules of failing models"
```
jochemvandooren, line 102: A PR has been merged recently to include the linting of sources, could you rename `models` to `items` everywhere? Then it's aligned!

src/dbt_score/formatters/human_readable_formatter.py
```diff
@@ -44,26 +44,34 @@ def evaluable_evaluated(
         """Callback when an evaluable item has been evaluated."""
         if score.value < self._config.fail_any_item_under:
             self._failed_evaluables.append((evaluable, score))
+        if score.value < self._config.fail_any_item_under or self._config.show in [
+            "all",
+            "failing-rules",
+        ]:
```
jochemvandooren, line 50: Currently, when linting my dbt project, the headers of models that pass are still shown when using `failing-rules`. e.g. 🥇 M: model_name (score: 10.0), without the rules.

So we have to change the if statements a bit!

src/dbt_score/formatters/human_readable_formatter.py
```diff
@@ -44,26 +44,34 @@ def evaluable_evaluated(
         """Callback when an evaluable item has been evaluated."""
         if score.value < self._config.fail_any_item_under:
             self._failed_evaluables.append((evaluable, score))
+        if score.value < self._config.fail_any_item_under or self._config.show in [
```
jochemvandooren, line 47: We can save `score.value < self._config.fail_any_item_under` to a param like `evaluable_failed`, to re-use in multiple places

tests/formatters/test_human_readable_formatter.py
```diff
@@ -28,6 +28,36 @@ def test_human_readable_formatter_model(
...
     expected = """\
+    🥇 \x1B[1mM: model1\x1B[0m (score: 10.0)
+        \x1B[1;31mERR \x1B[0m tests.conftest.rule_severity_medium: Oh noes
+        \x1B[1;33mWARN\x1B[0m (critical) tests.conftest.rule_severity_critical: Error
+
+    """
+    assert stdout == dedent(expected)
+
+
+def test_human_readable_formatter_model_show_all(
```
jochemvandooren, line 39: Can we have a test for `failing-models` as well?

jochemvandooren (commented):
Looks good, some minor comment on the output and some naming! :rocket:

src/dbt_score/cli.py
```diff
@@ -94,8 +94,20 @@ def cli() -> None:
...
+    "`failing-rules` shows failing rules of all models"
+    "Default behavior is to show violated rules of all models.",
```
matthieucan, line 104:
```suggestion
    "Default is --failing-rules.",
```
This is clearer IMO, and avoids parsing 2 sentences with the same meaning :)

src/dbt_score/formatters/human_readable_formatter.py
thomend: done:)

src/dbt_score/formatters/human_readable_formatter.py
thomend: you are right. I changed the if condition and tested it again with `tests/resources/manifest.json` . Now it does not print the evaluable's header if not failing and no failing rules.

docs/index.md
```diff
@@ -22,7 +22,8 @@ Score: 10.0 🥇
 
 In this example, the model `customers` scores the maximum value of `10.0` as it
 passes all the rules. It also is awarded a golden medal because of the perfect
-score.
+score. By default a passing model with or without rule violations will not be
```
jochemvandooren, line 25:
```suggestion
score. By default a passing model or source with or without rule violations will not be
```

jochemvandooren (commented):
Code looks good! :raised_hands: 

Happy to approve once you a test for the `--show failing-items` configuration, also please add a `CHANGELOG.md` entry! :rocket:

src/dbt_score/formatters/human_readable_formatter.py
jochemvandooren: Nice, thanks!

thomend:
Thanks for the review ! Hope the PR is now as you imagined.  
Just let me know if there is anything else outstanding on the PR :)

jochemvandooren (approved):
Awesome, thanks a lot! :rocket: Just need to add a `CHANGELOG.md` entry, but approving already!

thomend:
Sorry!  The changelog commit was not pushed! (pushed to remote right now)

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to
 ## [Unreleased]
 
 - Documenting support for python 3.13. (#86)
+- Only show failing rules per default in `HumanReadableFormatter`. Also added
+  `show` parameter in the CLI to change this behavior. (#77)
```
matthieucan, line 13:
```suggestion
  `--show` parameter in the CLI to change this behavior. (#77)
```

matthieucan (approved):
Thanks for your work @thomend! 😍 
Only formatting issues to fix with `prettier`, and it can be merged

thomend:
> Thanks for your work @thomend! 😍 Only formatting issues to fix with `prettier`, and it can be merged

Thank you as well! Yeah I saw it (it's odd though since I thought it runs also during pre-commit hooks). Will check it out.

jochemvandooren:
Awesome @thomend ! Thanks a lot for the contribution, major feature :star_struck: We will try to release it soon in `0.9.0`

thomend:
> Awesome @thomend ! Thanks a lot for the contribution, major feature 🤩 We will try to release it soon in `0.9.0`

Was very fun to work on and thank you for the opportunity!

PR #78, opened 2024-10-10
feat: lint sources

### LINKED ISSUE #76: feature request: enable linting of sources
This is a super cool tool!

Is it on the roadmap to implement dbt sources as a lint-able node type?

I'd be happy to contribute a PR if that's something worthwhile to this project.

otosky:
# Overview
Extends rules so that they can be used to against dbt [sources](https://docs.getdbt.com/docs/build/sources) in addition to models.

# Usage
A rule defines what resource-type it acts against in the type signature of the function it wraps or in a class-based `evaluate` method:

```python
from dbt_score import Model, Source rule, Rule, RuleViolation

# decorator-based
# for a Model
@rule
def model_has_description(model: Model) -> RuleViolation | None:
    """A model should have a description."""
    if not model.description:
        return RuleViolation(message="Model lacks a description.")

# for a Source
@rule
def has_description(source: Source) -> RuleViolation | None:
    """A source should have a loader defined."""
    if not source.loader:
        return RuleViolation(message="Source lacks a loader.")

# class-based
class ExampleSource(Rule):
    """Example class-based rule."""

    description = "A source should have a loader defined."

    def evaluate(self, source: Source) -> RuleViolation | None:
        """Evaluate source."""
        if not source.loader:
            return RuleViolation(message="Source lacks a loader.")
```

The `Evaluation` handler is then responsible for applying source-rules to Source objects and model-rules to Model objects.

---

closes #76


otosky:
woops thought that I opened this PR against my own fork - will move out of draft when I've added the rest of the feature :slightly_smiling_face:

matthieucan:
> woops thought that I opened this PR against my own fork - will move out of draft when I've added the rest of the feature 🙂

No problem, sounds good!

src/dbt_score/rule.py
```diff
@@ -78,6 +81,25 @@ def __init_subclass__(cls, **kwargs) -> None:  # type: ignore
...
+        sig = inspect.signature(evaluate_func)
+        resource_type_argument = first_true(
+            sig.parameters.values(),
+            pred=lambda arg: arg.annotation in typing.get_args(Evaluable),
+        )
+
+        if not resource_type_argument:
+            raise TypeError(
+                "Subclass must implement method `evaluate` with an annotated Model or Source argument."
+            )
```
otosky, line 99: Since the type-annotation is critical to a proper resource-type discriminator being set on the class, I thought that it was valid to throw an error here if there isn't a param typed as `Source` or `Model`.

otosky:
@matthieucan - let me know if you think this is heading in the right direction

src/dbt_score/evaluation.py
```diff
@@ -41,10 +43,10 @@ def __init__(
         self._scorer = scorer
 
         # For each model, its results
```
jochemvandooren, line 45:
```suggestion
        # For each evaluable, its results
```

src/dbt_score/evaluation.py
```diff
@@ -41,10 +43,10 @@ def __init__(
...
-        self.results: dict[Model, ModelResultsType] = {}
+        self.results: dict[Evaluable, EvaluableResultsType] = {}
 
         # For each model, its computed score
```
jochemvandooren, line 48:
```suggestion
        # For each evaluable, its computed score
```

src/dbt_score/evaluation.py
```diff
@@ -53,11 +55,13 @@ def evaluate(self) -> None:
         """Evaluate all rules."""
         rules = self._rule_registry.rules.values()
 
-        for model in self._manifest_loader.models:
+        for model in chain(self._manifest_loader.models, self._manifest_loader.sources):
```
jochemvandooren, line 58:
```suggestion
        for evaluable in chain(self._manifest_loader.models, self._manifest_loader.sources):
```

jochemvandooren: Also for further occurrences in this function, let's try to get rid of all the `model` references where now `sources` are included

src/dbt_score/__init__.py
```diff
@@ -1,11 +1,12 @@
 """Init dbt_score package."""
 
 from dbt_score.model_filter import ModelFilter, model_filter
```
jochemvandooren, line 3: We have to adapt `dbt_score.model_filter` in a similar way as `dbt_score.rule` was changed in order to apply filters to sources as well

jochemvandooren (commented):
I think this looks very good already! :rocket: Just a couple of things we need to think about:

- Filters cannot be applied to sources now
- Code has `model` references almost everywhere (oops :sweat_smile:, that was me probably), we should change that. E.g. `scorer.py` even has methods like `score_model` that will now be applied to `sources` as well.
- Can we distinguish source and model in the formatters?
- Needs some documentation in `/docs`!

Let me know what you think about it, I am happy to help if needed!

src/dbt_score/__init__.py
otosky: Updated in fe80bdf!  I will update the argument name references here as well in another commit.

How would you like to address the naming of the class/decorator now that it applies to both `Model`s and `Source`s?  Is it okay to have a breaking change here to rename it `RuleFilter` or `EvaluableFilter` or something else?

otosky:
> * Can we distinguish source and model in the formatters?

@jochemvandooren 

For the human-readable formatter, my thought was to prefix models with `M:` and sources with `S:`, but let me know if you have a different idea.

Example:
```
🥇 M:model1 (score: 10.0)
    OK   tests.conftest.rule_severity_low
    ERR  tests.conftest.rule_severity_medium: Oh noes
    WARN (critical) tests.conftest.rule_severity_critical: Error

🥇 S:source1.table1 (score: 10.0) 
    OK   tests.conftest.rule_severity_low
    ERR  tests.conftest.rule_severity_medium: Oh noes
    WARN (critical) tests.conftest.rule_severity_critical: Error
```

src/dbt_score/__init__.py
jochemvandooren: Nice! :raised_hands: I will have a look at the changes. 

I am afraid that this indeed warrants a breaking change, the name does not make sense anymore. But, breaking changes are expected considering we are still at version `0.6`. I like `RuleFilter` the most!

jochemvandooren:
> > * Can we distinguish source and model in the formatters?
> 
> @jochemvandooren
> 
> For the human-readable formatter, my thought was to prefix models with `M:` and sources with `S:`, but let me know if you have a different idea.
> 
> Example:
> 
> ```
> 🥇 M:model1 (score: 10.0)
>     OK   tests.conftest.rule_severity_low
>     ERR  tests.conftest.rule_severity_medium: Oh noes
>     WARN (critical) tests.conftest.rule_severity_critical: Error
> 
> 🥇 S:source1.table1 (score: 10.0) 
>     OK   tests.conftest.rule_severity_low
>     ERR  tests.conftest.rule_severity_medium: Oh noes
>     WARN (critical) tests.conftest.rule_severity_critical: Error
> ```

I like the idea, keeps it concise! :+1:

src/dbt_score/models.py
```diff
@@ -266,6 +403,13 @@ def _load_models(self) -> None:
...
                 self.models.append(model)
 
+    def _load_sources(self) -> None:
+        """Load the sources from the manifest."""
+        for source_id, source_values in self.raw_sources.items():
+            if source_values.get("resource_type") == "source":
+                source = Source.from_node(source_values, self.tests.get(source_id, []))
+                self.sources.append(source)
+
     def _reindex_tests(self) -> None:
```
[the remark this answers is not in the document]
jochemvandooren: Same for `_select_models`

jochemvandooren (commented):
This has become a huge PR with all the renaming and stuff, great effort :raised_hands: I have played around with the code locally and everything works perfectly! Just left some comments about getting rid of all mentions of `model` in the code.

Also, all of the documentation needs to be updated as well, I am happy to assist with this one, please let me know!

src/dbt_score/cli.py
```diff
@@ -87,8 +87,8 @@ def cli() -> None:
     default=None,
 )
 @click.option(
-    "--fail_any_model_under",
-    help="Fail if any model is under this value.",
+    "--fail_any_evaluable_under",
+    help="Fail if any evaluable is under this value.",
```
otosky, line 91: @jochemvandooren 

Were you okay with this change in configuration?  I think this would be my preference (a single setting to fail any item under a threshold), but I could see the argument for separate settings for models or sources as well.

otosky:
Thanks for a thorough review @jochemvandooren !  Will address remaining feedback, update the formatters, and take a first stab at some of the docs shortly!  I will definitely take up your offer for support on the docs.

src/dbt_score/cli.py
matthieucan: While there, could we use dashes `-` instead of underscores `_` in the CLI option? This was a mistake when introduced 🙏

matthieucan: "evaluable" seems very correct and fancy, but perhaps not the best for the CLI. Maybe "entity" could work better, as in dbt entities are models, sources, etc. WDYT?

src/dbt_score/models.py
matthieucan, line 1: Should this file be renamed accordingly to something like `dbt_entities.py`?

matthieucan (commented):
Incredible work @otosky !
I played a bit with it and it works fine! Very much looking forward to have this 💪

src/dbt_score/models.py
```diff
@@ -247,11 +376,19 @@ def __init__(self, file_path: Path, select: Iterable[str] | None = None):
...
+            if source_values["package_name"] == self.project_name
+        }
+
         self.models: list[Model] = []
         self.tests: dict[str, list[dict[str, Any]]] = defaultdict(list)
+        self.sources: list[Source] = []
 
         self._reindex_tests()
         self._load_models()
+        self._load_sources()
```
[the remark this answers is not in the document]
otosky: Updated!

src/dbt_score/models.py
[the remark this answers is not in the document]
otosky: Updated `_select_models` -> `_filter_evaluables`.

Note that the selector name output by `dbt ls --output name` for sources is of the form `source_name.source_table_alias`.

FWIW, I have some thoughts for another leg of work to add some better integration testing here - all the mocking/stubbing in the manifest related codepaths didn't leave me feeling confident that the changes I was making would actually work on a real project.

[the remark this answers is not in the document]
otosky: Updated `_reindex_tests` - it needed a little massaging to fit sources.

src/dbt_score/rule_filter.py
```diff
@@ -41,31 +64,31 @@ def __hash__(self) -> int:
```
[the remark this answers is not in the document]
otosky: updated

src/dbt_score/evaluation.py
```diff
@@ -53,26 +54,32 @@ def evaluate(self) -> None:
...
+        for evaluable in chain(
+            self._manifest_loader.models, self._manifest_loader.sources
+        ):
+            self.results[evaluable] = {}
             for rule in rules:
                 try:
-                    if rule.should_evaluate(model):  #  Consider model filter(s).
-                        result = rule.evaluate(model, **rule.config)
-                        self.results[model][rule.__class__] = result
+                    if rule.resource_type is type(evaluable) and rule.should_evaluate(
```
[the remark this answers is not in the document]
otosky: This is actually a different check to match up the resource_type of the rule to the resource_type of the model.  But based on that confusion, I think it's probably better to push that down into the `rule.should_evaluate` method?

src/dbt_score/cli.py
otosky: updated

src/dbt_score/evaluation.py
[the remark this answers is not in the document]
otosky: updated in 9b5af62

src/dbt_score/models.py
otosky: IMO `models` is idiomatic enough for it's contents that I don't think users will get too tripped up with the mixed terminology.  I don't feel strongly either way, though!

docs/create_rules.md
```diff
@@ -91,30 +113,48 @@ def sql_has_reasonable_number_of_lines(model: Model, max_lines: int = 200) -> Ru
...
     return model.schema.lower() == 'x'
 
-class SkipSchemaY(ModelFilter):
+class SkipSchemaY(RuleFilter):
     description = "Applies a rule to every schema but Y."
     def evaluate(self, model: Model) -> bool:
       return model.schema.lower() != 'y'
 ```
 
+Filters also rely on type-annotations to dictate whether they apply to models or sources:
```
otosky, line 141: I'm realizing now that I haven't actually made any changes to enforce that resource_types match across `rule_filter`s attached to `rule`s.  

This would be easy to account for by addressing it in the `rule.should_evaluate` method, but would it be better to throw an error earlier with a nice message such that something like this can't be configured at all?

```python
    @rule_filter
    def source_schema_in_x(source: Source):
        """Only keep sources in schema x."""
        return schema_in_x(source)

    @rule(rule_filters={source_schema_in_x()})
    def model_has_description(model: Model):
        """Validates that a model has a description."""
        if not model.description:
            return RuleViolation("Model does not have a description.")
```

src/dbt_score/models.py
jochemvandooren: I came to the same conclusion @otosky , no strong opinion either!

docs/create_rules.md
jochemvandooren: I agree throwing an error earlier makes more sense indeed, it improves UX in my opinion! :+1:

jochemvandooren:
> Thanks for a thorough review @jochemvandooren ! Will address remaining feedback, update the formatters, and take a first stab at some of the docs shortly! I will definitely take up your offer for support on the docs.

Next week I will help you on the docs and further review the PR, thanks for all the amazing work already :raised_hands:

src/dbt_score/models.py
matthieucan: Alright! 👍

docs/create_rules.md
otosky: addressed in 0eeae68

src/dbt_score/evaluation.py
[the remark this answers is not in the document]
jochemvandooren: Nice, thanks!

docs/create_rules.md
jochemvandooren: Great :raised_hands:

jochemvandooren:
> > Thanks for a thorough review @jochemvandooren ! Will address remaining feedback, update the formatters, and take a first stab at some of the docs shortly! I will definitely take up your offer for support on the docs.
> 
> Next week I will help you on the docs and further review the PR, thanks for all the amazing work already 🙌

@otosky The code looks perfect! As promised I did a final check and tried to find all occurrences of `model` and replaced it by something more appropriate :mag: https://github.com/PicnicSupermarket/dbt-score/pull/78/commits/d595ea278881b062ca5613b6fd3d190ecc048fab

I also added a `CHANGELOG.md` entry already, feel free to improve of course. Final thing that's remaining is some linting errors on lines exceeding line length. Once those are fixed I am ready to approve :rocket:

otosky:
Thank you @jochemvandooren!  

Fixed the line lengths in a2c0924 and made some tweaks to the changelog in f88aa7c.

One final Q:  I made use of two functions from [more-itertools](https://more-itertools.readthedocs.io/en/stable/) - [first](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.first) and [first_true](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.first_true).  `more-itertools` came as a sub-dependency of `dbt`, which I see has now been made a dev-dep.  Would you like me to rewrite/vendor the 2 functions I used above?  Or add `more-itertools` as proper dependency?

jochemvandooren:
> Thank you @jochemvandooren!
> 
> Fixed the line lengths in [a2c0924](https://github.com/PicnicSupermarket/dbt-score/commit/a2c09242ec2e8e80ee85c37e6e87d80c7cd758bd) and made some tweaks to the changelog in [f88aa7c](https://github.com/PicnicSupermarket/dbt-score/commit/f88aa7c249ee7b0e6b031c80c331bd6617658a94).
> 
> One final Q: I made use of two functions from [more-itertools](https://more-itertools.readthedocs.io/en/stable/) - [first](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.first) and [first_true](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.first_true). `more-itertools` came as a sub-dependency of `dbt`, which I see has now been made a dev-dep. Would you like me to rewrite/vendor the 2 functions I used above? Or add `more-itertools` as proper dependency?

Ah, good point! If it can be prevented, I'd like to keep the number of dependencies low. But I can imagine having to rewrite the function is a hassle, I will leave it up to you!

otosky:
I went the vendoring route, since it's really just the function `first_true` that is nicer syntax sugar and used in more than one place.  :+1:

matthieucan:
> I went the vendoring route, since it's really just the function `first_true` that is nicer syntax sugar and used in more than one place. 👍

I believe `more_itertools.py` was not committed?

otosky:
> > I went the vendoring route, since it's really just the function `first_true` that is nicer syntax sugar and used in more than one place. 👍
> 
> I believe `more_itertools.py` was not committed?

:facepalm: totally right - just pushed it up!

jochemvandooren (approved):
Amazing :star_struck: , just needs a rebase on master!

CHANGELOG.md
```diff
@@ -8,6 +8,17 @@ and this project adheres to
 
 ## [Unreleased]
 
+- **Breaking**: Support linting of sources.
```
matthieucan, line 11: That's a feature, not a breaking change, right?

src/dbt_score/rule_filter.py
```diff
@@ -1,16 +1,19 @@
-"""Model filtering to choose when to apply specific rules."""
-
+"""Evaluable filtering to choose when to apply specific rules."""
```
matthieucan, line 1:
```suggestion
"""Evaluable filtering to choose when to apply specific rules."""

```

matthieucan: (adding empty line)

matthieucan (approved):
Minor nitpicks.

Incredible work @otosky, I'm really keen to start using this feature. Thanks a lot! 💪

CHANGELOG.md
otosky: Yeah, I'll fix that

src/dbt_score/formatters/manifest_formatter.py
```diff
@@ -15,19 +15,19 @@ class ManifestFormatter(Formatter):
...
+        self._evaluable_scores[evaluable.unique_id] = score
 
     def project_evaluated(self, score: Score) -> None:
         """Callback when a project has been evaluated."""
         manifest = copy.copy(self._manifest_loader.raw_manifest)
-        for model_id, model_score in self._model_scores.items():
-            manifest["nodes"][model_id]["meta"]["score"] = model_score.value
-            manifest["nodes"][model_id]["meta"]["badge"] = model_score.badge
+        for evaluable_id, evaluable_score in self._evaluable_scores.items():
+            manifest["nodes"][evaluable_id]["meta"]["score"] = evaluable_score.value
```
otosky, line 31: Noticing after taking another pass at the diff that sources need to be pushed into `manifest["sources"]` instead of `manifest["nodes"]`.  Let me fix that quickly.

matthieucan: Oh good one, we should have a test for that

otosky: updated in cc6b707

src/dbt_score/formatters/manifest_formatter.py
```diff
@@ -15,19 +15,25 @@ class ManifestFormatter(Formatter):
...
-            manifest["nodes"][model_id]["meta"]["badge"] = model_score.badge
+        for evaluable_id, evaluable_score in self._evaluable_scores.items():
+            if evaluable_id.startswith("model"):
+                model_manifest = manifest["nodes"][evaluable_id]
+                model_manifest["meta"]["score"] = evaluable_score.value
+                model_manifest["meta"]["badge"] = evaluable_score.badge
+            if evaluable_id.startswith("source"):
+                source_manifest = manifest["sources"][evaluable_id]
+                source_manifest["meta"]["score"] = evaluable_score.value
+                source_manifest["meta"]["badge"] = evaluable_score.badge
```
otosky, line 38: I did the simple thing here and didn't account for any future situation where an Evaluable is something outside of Model | Source, but let me know if it's better to be more defensive here.  Open to suggestions.

jochemvandooren: I think fine for now! :ok_hand:

jochemvandooren:
It seems `mypy` was upgraded when rebasing, and it introduced some new linting errors :sweat_smile: I can help you fix them if you would like! Let me know

otosky:
@jochemvandooren I'll take an initial stab at it!

otosky:
There are still a bunch of Liskov violations being raised from mypy that I'm not entirely sure how to fix without having to introduce Generics into the API.

```
tests/test_rule.py:53: error: Argument 1 of "evaluate" is incompatible with supertype "Rule"; supertype defines the argument type as "Model | Source"  [override]
tests/test_rule.py:53: note: This violates the Liskov substitution principle
tests/test_rule.py:53: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

Is there any other workaround besides adding an ignore here?  This essentially means that mypy will fail for any downstream users if they use the class-based workflow while developing their rules.

jochemvandooren:
> There are still a bunch of Liskov violations being raised from mypy that I'm not entirely sure how to fix without having to introduce Generics into the API.
> 
> ```
> tests/test_rule.py:53: error: Argument 1 of "evaluate" is incompatible with supertype "Rule"; supertype defines the argument type as "Model | Source"  [override]
> tests/test_rule.py:53: note: This violates the Liskov substitution principle
> tests/test_rule.py:53: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
> ```
> 
> Is there any other workaround besides adding an ignore here? This essentially means that mypy will fail for any downstream users if they use the class-based workflow while developing their rules.

I also spent some time looking into this, and there's no easy solution indeed :disappointed: Considering this will only affect the class-based rules, I suggest we add an ignore for now. I am aware it will introduce the warnings downstream, which isn't ideal. To solve this in a nice way, we might need to restructure some things if we want to keep the API as is, so we might consider this in a follow-up PR. What do you think? I think it's the most pragmatic option!

otosky:
> I also spent some time looking into this, and there's no easy solution indeed 😞 Considering this will only affect the class-based rules, I suggest we add an ignore for now. I am aware it will introduce the warnings downstream, which isn't ideal. To solve this in a nice way, we might need to restructure some things if we want to keep the API as is, so we might consider this in a follow-up PR. What do you think? I think it's the most pragmatic option!

makes sense to me @jochemvandooren! - updated in 0732e15

otosky:
@jochemvandooren @matthieucan thanks again for all your assistance!

PR #80, opened 2024-10-22
Remove filter in rule public_model_has_example_sql

matthieucan:
A rule should not care about public/private models, for this a RuleFilter should be used.


jochemvandooren (approved):
Agreed!

druzhinin-kirill (approved):

PR #81, opened 2024-10-28
Remove dbt-core dependency

jochemvandooren:
Get rid of the dependency `dbt-core`.

In order to keep the number of dependencies as low as possible, to keep python environments small, get rid of the dbt-core dependency and add it to the dev dependencies. This is possible because `dbt-score` only needs a `manifest.json` in order to run, although `dbt-score` is able to run `dbt parse` by itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to
 
 ## [Unreleased]
 
+- **Breaking**: The rule `public_model_has_example_sql` has been renamed
+  `has_example_sql` and applies by default to all models.
 - **Breaking**: Remove `dbt-core` from dependencies. Since it is not mandatory
   for `dbt-score` to execute `dbt`, remove the dependency.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to
 
 ## [Unreleased]
 
+- **Breaking**: Remove `dbt-core` from dependencies. Since it is not mandatory
+  for `dbt-score` to execute `dbt`, remove the dependency.
+
 ## [0.6.0] - 2024-08-23
 
 - **Breaking**: Improve error handling in CLI. Log messages are written in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
   `has_example_sql` and applies by default to all models.
 - **Breaking**: Remove `dbt-core` from dependencies. Since it is not mandatory
   for `dbt-score` to execute `dbt`, remove the dependency.
+- **Breaking**: Stop using `MultiOption` selection type.
 
 ## [0.6.0] - 2024-08-23
 
--- a/docs/get_started.md
+++ b/docs/get_started.md
@@ -12,8 +12,7 @@ pip install dbt-score
 ```
 
 In order to run `dbt-score` with all its features, be sure to install
-`dbt-score` in the same environment as `dbt-core` as some features require
-`dbt-core` to be installed.
+`dbt-score` in the same environment as `dbt-core`.
 
 ## Usage
 
--- a/docs/get_started.md
+++ b/docs/get_started.md
@@ -11,7 +11,7 @@ Installation of `dbt-score` is simple:
 pip install dbt-score
 ```
 
-In order to run `dbt-score` with all it's features, ensure to install
+In order to run `dbt-score` with all its features, be sure to install
 `dbt-score` in the same environment as `dbt-core` as some features require
 `dbt-core` to be installed.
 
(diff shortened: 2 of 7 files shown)


src/dbt_score/cli.py
```diff
@@ -48,9 +53,10 @@ def cli() -> None:
     "--select",
     "-s",
     help="Specify the nodes to include.",
-    cls=MultiOption,
+    cls=MultiOption if DBT_INSTALLED else None,
```
matthieucan, line 56: Makes me wonder if we should remove that feature, and enforce `--select foo --select bar` instead of `--select foo bar`.

src/dbt_score/dbt_utils.py
```diff
@@ -31,8 +42,13 @@ def dbt_parse() -> dbtRunnerResult:
     Raises:
         DbtParseException: dbt parse failed.
     """
+    if not DBT_INSTALLED:
+        raise DbtNotInstalledException(
+            "To execute dbt parse, install dbt-core in your environment."
```
matthieucan, line 47:
```suggestion
            "This option requires dbt to be installed in the same Python environment as dbt-score."
```
Maybe this phrasing?

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to
 
 ## [Unreleased]
 
+- **Breaking**: Remove `dbt-core` from dependencies. Since it is not mandatory
+  for `dbt-score` to execute `dbt`, remove the dependency.
```
matthieucan, line 12:
```suggestion
- **Breaking**: The rule `public_model_has_example_sql` has been renamed
  `has_example_sql` and applies by default to all models.
- **Breaking**: Remove `dbt-core` from dependencies. Since it is not mandatory
  for `dbt-score` to execute `dbt`, remove the dependency.
```
While there, I did forget to document this change 😇

jochemvandooren: Ah good one :+1:

src/dbt_score/cli.py
jochemvandooren: To be honest I like how we have the exact same select statement that `dbt` uses, it will prevent any confusion on how users are able to select models (and sources) in the future

src/dbt_score/dbt_utils.py
jochemvandooren: I like it!

pyproject.toml
```diff
@@ -23,7 +23,6 @@ classifiers = [
 ]
 
 dependencies = [
-    "dbt-core>=1.5",
```
druzhinin-kirill, line 26: As far as we still have functionality within a package that uses `dbt-core` (conditionally), did you consider making it a dev dependency?

src/dbt_score/dbt_utils.py
druzhinin-kirill, line 47: Not sure if we plan to have more features requiring dbt, but if so, we can also consider making a decorator `@dbt_required` and implement this check to avoid repetitions. 
However, with 2 occurrences now seems to be to much

pyproject.toml
jochemvandooren: Good idea, added it to the dev dependencies 👍

docs/get_started.md
```diff
@@ -11,8 +11,9 @@ Installation of `dbt-score` is simple:
 pip install dbt-score
 ```
 
-If a virtual environment is used to run dbt, make sure to install `dbt-score` in
-the same environment.
+In order to run `dbt-score` with all it's features, ensure to install
```
matthieucan, line 14:
```suggestion
In order to run `dbt-score` with all its features, be sure to install
```

docs/get_started.md
```diff
@@ -11,8 +11,9 @@ Installation of `dbt-score` is simple:
...
+`dbt-score` in the same environment as `dbt-core` as some features require
+`dbt-core` to be installed.
```
matthieucan, line 16:
```suggestion
`dbt-score` in the same environment as `dbt-core`.
```
Simplify sentence

matthieucan (approved):

src/dbt_score/dbt_utils.py
```diff
@@ -16,13 +25,27 @@ class DbtLsException(Exception):
     """Raised when dbt ls fails."""
 
 
+def dbt_required(func: Callable[..., Any]) -> Callable[..., Any]:
+    """Decorator for methods that require dbt to be installed."""
+
+    def inner() -> None:
```
druzhinin-kirill, line 31: We can use https://docs.python.org/3/library/functools.html#functools.wraps

to preserve docstrings, names etc of the original function 👌

druzhinin-kirill (approved):
Left one comment, PHAL, but preapproving anyways 👌

PR #87, opened 2024-12-13
Prevent duplicated rule exception when importing rules and filters

### LINKED ISSUE #85: Bug: Importing a definition is counted by rule registry
When a rule or filter is defined **and** imported in another file that are both in the `rule_namespaces` directories, the object will be counted twice, resulting in an error.

### How to reproduce
Let's assume rules and filters are stored in `dbt_score_rules`.

- Create a filter in `dbt_score_rules/rule_filters.py`
- Import the filter in `dbt_score_rules/model_rules.py`
- Run `dbt-score list`, to list all rules
- Returns: `dbt_score.exceptions.DuplicatedRuleException: Rule dbt_score_rules.rule_filtes.is_edge_model is defined twice. Rules must have unique names.`

npeshkov:
Fixes #85.

### Implementation

- Skip adding imported rules and rule filters to rule registry.
- Split `tests/rules/example.py` into `tests/rules/rules.py` and `tests/rules/rule_filters.py`.
- Test `Duplicated` exception is not raised by adding an imported `skip_schemaX` rule filter.

### Note

Both rules and rule filters can be imported now, but the existing tests cover only rule filters.


src/dbt_score/rule_registry.py
```diff
@@ -62,6 +62,9 @@ def _load(self, namespace_name: str) -> None:
             module = importlib.import_module(module_name)
             for obj_name in dir(module):
                 obj = module.__dict__[obj_name]
+                # skip adding objects imported from other modules
```
jochemvandooren, line 65:
```suggestion
                # Skip adding objects imported from other modules
```
:nerd_face:

npeshkov: Thanks, should be fixed now.

tests/rules/imported/__init__.py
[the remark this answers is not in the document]
npeshkov: Good point, I have split `rules/example.py` into `rules.py` and `rule_filters.py`, and added a rule filter which is now imported into `rules.py`. The structure should look better now.

jochemvandooren (approved):
Nice! :star_struck:

jochemvandooren:
I will give @matthieucan the final say as he worked on the rule registry

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to
 ## [Unreleased]
 
 - Documenting support for python 3.13. (#86)
+- Ignore imported rules and filters when building the rule registry. (#85)
```
matthieucan, line 12:
```suggestion
- Ignore imported rules and filters when building the rule registry. (#87)
```

matthieucan: Let's refer to the current PR, instead of the bug report. This allows e.g. PRs to partially solve bug reports, and therefore the changelog to be more precise :)

matthieucan (approved):
Tiny comment, but approving. Thanks!

PR #90, opened 2024-12-24
Add generic rule: models should implement uniqueness test for their PK

matthieucan:
For models materialized as `table` or `incremental`:
- Loop over their columns to extract PK
- Loop over their tests to find a uniqueness test matching the PK columns

Consider both table- and column-level constraints and tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to
 - Add debug mode to help writing new rules. (#91)
 - Fix tests without metadata. (#88)
 - Add new rule to enforce presence of uniqueness test. (#90)
+- Add `constraints` to the model schema. (#90)
 
 ## [0.9.0] - 2024-12-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
-this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to
+[Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
@@ -11,14 +12,15 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix tests without metadata. (#88)
 - Add new rule to enforce presence of uniqueness test. (#90)
 - Add new rule to enforce single-column PK to be defined at column level. (#90)
-- Add new rule to enforce 1 column uniqueness test to be defined at column level. (#90)
+- Add new rule to enforce 1 column uniqueness test to be defined at column
+  level. (#90)
 - Add `constraints` to the model schema. (#90)
 
 ## [0.9.0] - 2024-12-19
 
 - Documenting support for python 3.13. (#86)
-- Only show failing rules per default in `HumanReadableFormatter`. Also added `--show`
-  parameter in the CLI to change this behavior. (#77)
+- Only show failing rules per default in `HumanReadableFormatter`. Also added
+  `--show` parameter in the CLI to change this behavior. (#77)
 - Ignore imported rules and filters when building the rule registry. (#87)
 
 ## [0.8.0] - 2024-11-12
@@ -30,7 +32,8 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   `@rule_filter` and `ModelFilter` becomes `RuleFilter`.
 - **Breaking**: Config option `model_filter_names` becomes `rule_filter_names`.
 - **Breaking**: CLI flag naming fixes: `--fail_any_model_under` becomes
-  `--fail-any-item-under` and `--fail_project_under` becomes `--fail-project-under`.
+  `--fail-any-item-under` and `--fail_project_under` becomes
+  `--fail-project-under`.
 
 ## [0.7.1] - 2024-11-01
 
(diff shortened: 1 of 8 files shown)


src/dbt_score/rules/generic.py
```diff
@@ -47,7 +47,49 @@ def sql_has_reasonable_number_of_lines(
...
+        if "```sql" not in (model.description or ""):
             return RuleViolation(
                 "The model description does not include an example SQL query."
             )
+
+@rule
+def has_uniqueness_test(model: Model) -> RuleViolation | None:
+    """Model has uniqueness test for primary key."""
+    if model.config.get("materialized") not in {"table", "incremental"}:
+        return None
```
druzhinin-kirill, line 59: Would you like us to implement this with a rule filter? Maybe even the whole pk discovery can be a filter, and this filter can be also part of the public dbt-score API on par with default rules 🤔 

So people who implement their own rules but use the same constraints on filters can reuse it 💡

src/dbt_score/rules/generic.py
```diff
@@ -47,7 +47,49 @@ def sql_has_reasonable_number_of_lines(
...
+
+    # Extract PK
+    pk_columns = None
```
druzhinin-kirill, line 62: Can it be an empty list, and we only verify pk_columns if the length is >0?
I think this way we can ensure `pk_columns` is always a list

druzhinin-kirill (commented):
💪

src/dbt_score/rules/generic.py
matthieucan: That's a great idea! Unfortunately the logic to extract the PK can't be in the filter, but that works nonetheless. Added!

src/dbt_score/rules/generic.py
matthieucan: I think I prefer it to be `None`, to have the distinction between "we haven't found a PK" and "there's an empty PK" - even though the latter is not a possibility, I believe it makes the code more readable

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to
 
 - Add debug mode to help writing new rules. (#91)
 - Fix tests without metadata. (#88)
+- Add new rule to enforce presence of uniqueness test. (#90)
```
jochemvandooren, line 13: Let's document that we added `constraints` to the model properties

src/dbt_score/models.py
```diff
@@ -187,6 +187,7 @@ class Model(HasColumnsMixin):
     config: dict[str, Any]
     meta: dict[str, Any]
     columns: list[Column]
+    constraints: list[dict[str, Any]]
```
jochemvandooren, line 190: Can we use `Constraint`, like we do with columns as well?

jochemvandooren (commented):
Looks good! Some minor comments

src/dbt_score/models.py
matthieucan: Good point, done!

src/dbt_score/rules/generic.py
```diff
@@ -47,7 +48,49 @@ def sql_has_reasonable_number_of_lines(
...
+                for data_test in column.tests:
+                    if data_test.type == "unique":
+                        return None
+
+    for data_test in model.tests:
+        if data_test.type == "unique_combination_of_columns":
+            if set(data_test.kwargs.get("combination_of_columns")) == set(pk_columns): # type: ignore
+                return None
+
+    return RuleViolation("There is no uniqueness test defined and matching the PK.")
```
druzhinin-kirill, line 96:
```suggestion
    pk_columns = next(
        (
            [column.name]
            for column in model.columns
            for constraint in column.constraints
            if constraint.type == "primary_key"
        ),
        [],
    )

    if not pk_columns:
        pk_columns = next(
            (
                constraint.columns
                for constraint in model.constraints
                if constraint.columns and constraint.type == "primary_key"
            ),
            [],
        )

    if len(pk_columns) == 1:
        pk_column = next(
            column for column in model.columns if column.name == pk_columns[0]
        )
        if not any(test.type == "unique" for test in pk_column.tests):
            return RuleViolation("There is no uniqueness test defined.")

    elif len(pk_columns) > 1:
        constraint_test = next(
            (
                test
                for test in model.tests
                if test.type == "unique_combination_of_columns"
            ),
            None,
        )
        if not constraint_test:
            return RuleViolation("There is no uniqueness test defined.")
        if set(constraint_test.kwargs.get("combination_of_columns", [])) == set(
            pk_columns
        ):
            return RuleViolation("Uniqueness test do not match the PK.")
```

I tried to get rid of lint exceptions and make it a bit more flat - feel free to use fully or partially

matthieucan: Thanks for challenging! I adopted some of it, making the function flatter.
I didn't take the list comprehensions however, as keeping `break`s improves performances a bit. I could get rid of one ruff-ignore, not the other unfortunately.
Overall, while simplifying the logic, I was led to add 2 related rules, making it clearer when a model doesn't conform to expectations. PTAL :)

src/dbt_score/rules/generic.py
```diff
@@ -53,44 +53,60 @@ def has_example_sql(model: Model) -> RuleViolation | None:
...
+        return None
 
     for data_test in model.tests:
         if data_test.type == "unique_combination_of_columns":
             if set(data_test.kwargs.get("combination_of_columns")) == set(pk_columns): # type: ignore
                 return None
-
-    return RuleViolation("There is no uniqueness test defined and matching the PK.")
+    return RuleViolation(
+        f"No uniqueness test defined and matching PK {','.join(pk_columns)}."
```
npeshkov, line 111:
```suggestion
        f"No uniqueness test defined for matching Primary Keys {','.join(pk_columns)}."
```

src/dbt_score/rules/generic.py
```diff
@@ -53,44 +53,60 @@ def has_example_sql(model: Model) -> RuleViolation | None:
...
-
-    # Look for matching uniqueness test
-    if len(pk_columns) == 1:
-        for column in model.columns:
-            if column.name == pk_columns[0]:
                 for data_test in column.tests:
                     if data_test.type == "unique":
                         return None
+                return RuleViolation(
+                    f"No unique constraint defined on PK column {column.name}."
```
npeshkov, line 93:
```suggestion
                    f"No unique constraint defined on Primary Key column {column.name}."
```

npeshkov (approved):
Proposing some changes to text and approving 👍

src/dbt_score/rules/generic.py
```diff
@@ -53,44 +53,60 @@ def has_example_sql(model: Model) -> RuleViolation | None:
...
+        if data_test.type == "unique_combination_of_columns":
+            if len(data_test.kwargs.get("combination_of_columns")) == 1:
+                return RuleViolation(
+                    f"Single-column uniqueness test must be defined as a column test."
+                )
+
+
 @rule(rule_filters={is_table()})
 def has_uniqueness_test(model: Model) -> RuleViolation | None:
     """Model has uniqueness test for primary key."""
```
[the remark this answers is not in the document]
matthieucan: I could go with `has_uniqueness_test_for_pk`, but find the suggestion too verbose

src/dbt_score/rules/generic.py
matthieucan: While acronyms are sometimes more confusing than anything else, in this context, I believe PK is well understood

PR #101, opened 2025-02-25
Do not ignore docs/ changes in CI

matthieucan:
To optimize builds, changes affecting only the `docs/` folder were ignored in CI.
However, `prettier` has been added in the meantime to lint/format markdown files, which can lead to unformatted files making their way to the main branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,14 +4,10 @@ on:
   pull_request:
     branches:
       - master
-    paths-ignore:
-      - "docs/**"
 
   push:
     branches:
       - master
-    paths-ignore:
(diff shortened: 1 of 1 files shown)


druzhinin-kirill:
One possible option is to create a `docs-ci` that will only run when docs are touched and will only run related doc steps (prettier lint). However, this will also simply work 👍

matthieucan:
> One possible option is to create a `docs-ci` that will only run when docs are touched and will only run related doc steps (prettier lint). However, this will also simply work 👍

Nice, let's keep it in mind for a potential optimization :)

jochemvandooren (approved):

Issue #98, opened 2025-02-13
Refs, and rules about model lineage


ross-whatnot:
We have some use cases where we're interested in making assertions about model refs, in a couple different categories. 

The simpler case would be, like, just making assertions about the refs within a given model - how many refs does it have, etc? This seems pretty easily achievable by just surfacing the list of refs in the model to be used in rules.

The slightly more complex case would be to make assertions about the refs in a given model based on characteristics of the models _being ref'd_; like, does this model ref any models that are in schema x, or that have a given piece of metadata? That seems like it'd require a bit more lift, since the rule would need access to not just the specific model it's evaluating, but also to the _other_ models that exist in the graph.

The former seems like something I could take a crack at adding, but curious how the latter feels?

jochemvandooren:
Hey @ross-whatnot , thanks for opening an issue!

The first one seems doable: the `manifest.json` has some information of the refs in the `depends_on` attribute (if I remember correctly), so it's definitely something we should be able to add.

The second one is a little bit difficult at this point in time, and not sure how we would like to proceed with this. 🤔 Currently the scope of a `rule` is a `model` or `source`, on it's own. We do not have an option to look beyond the `model` or `source`. If we want to, we would have to make the full `manifest` available to the `rule`, which makes things definitely more complicated. Another option would be to add the full `Model` or `Source` to the `depends_on` property that we are looking to add in option one. This might be not _too_ complicated!

jochemvandooren:
Maybe we can start simple: adding the `depends_on` as strings (as they are saved in the manifest) and in a followup issue/PR try to find the `Model`/`Source` objects.

matthieucan:
Following on the dependency injection approach that dbt-score follows, would it be a possibility have a rule declare the need to access the full manifest in such a fashion:

```python
@rule
def my_rule(model: Model, manifest: Manifest):
    # do some assertions about `model` using data found in the `manifest`
```

Alternatively, `Manifest` could be the already existing `ManifestLoader` class.

ross-whatnot:
It looks like `depends_on` is already there, actually (and includes macros, which is nice)! I'll see what I can do with that, and in doing so think about what would be gained by having access to the model defs themselves. Just for context, the kind of rules I'm interested in here are kind of like some of what [dbt project evaluator](https://dbt-labs.github.io/dbt-project-evaluator/latest/rules/modeling/) does (assertions about the shape of the graph).

jochemvandooren:
[...]

PR #106, opened 2025-03-29
Sketch of providing access to manifest

ross-whatnot:
As mentioned in [this issue](https://github.com/PicnicSupermarket/dbt-score/issues/98), I have use cases where it's valuable to be able to consider characteristics of ref'd nodes in evaluating a given node. The simplest case version here would be something like:
* a model that has a tag `tier_1` may only depend on other models or sources that also have the tag `tier_1`
* a model in folder `silver` may only depend on other models that are in folder `silver` or `bronze`
* a model that does not have the tag `beta` may not depend on any model that does have the tag `beta` (or, alternatively, all children of a model that has the tag `beta` must also have the tag `beta`)

I put this together as a sketch of an approach that might satisfy this kind of testing need. After building it out, though, this feels somewhat un-ergonomic; hope that it provides a good jumping off point for discussion.
* My primary reflection is that within a given test, the only easy way to access dependencies is to look up FQNs from `depends_on` in the `raw_nodes` of the manifest; this works, but then you just have the raw dictionary, rather than a nice Model or Source object. This _can_ be beneficial, as it provides access to macros / other things that don't have a native representation, but it's clunky for the average case (of using a model or source). My inclination would be to either:
    * refactor this to provide a more native `parents: dict[str, Model | Source]` (and maybe `children`) dictionary (or just provide `parents: list[Model | Source]`?) that can be optionally requested by rules, in a similar manner to how this sketch shows optionally requesting the manifest
    * or, refactor this to put `parents` (and `children`) directly into the `Model` / `Source` models themselves, similar to how tests are handled now. This might be the most ergonomic in a lot of ways from an end user perspective, but could make each model quite a bit larger and might cause other headaches.
* From a code perspective, is there a cleaner way to say "provide this parameter if it's requested, otherwise don't"? I believe what I have here _works_, shouldn't require any changes to existing rule signatures, and feels reasonably in alignment with existing interfaces, but is a bit clunky (in particular, the usage of `requires_manifest()` in `evaluation.py` and associated awkwardness in tests). (If went with the "put it in the model" approach, this problem maybe goes away)
* Aside: Another half-formed thought here is that this remains very centered around considering a model and its immediate neighbors, rather than the graph as a whole. I think this feels _appropriate_; there are certainly more "graph shape"-oriented tests I can imagine, but they're rather different from the existing testing modalities in this project and feel reasonable to set aside (for now), I think?

Look forward to discussing!


jochemvandooren:
Hey @ross-whatnot, thanks for your great write-up! :raised_hands:

Looking at your code and after running it locally I can conclude it works as intended indeed, very nice. But yes, it feels a little bit clunky/un-ergonomic because we can only access raw values and not the `Model` or `Source` objects. I am in favor of keeping things as simple and clear as possible to the user, so probably navigating through the raw manifest is not the way to go here (assuming the average dbt-score rule-writer does not know what exact properties are present in the manifest).

> or, refactor this to put parents (and children) directly into the Model / Source models themselves, similar to how tests are handled now. This might be the most ergonomic in a lot of ways from an end user perspective, but could make each model quite a bit larger and might cause other headaches.

To me, this sounds like a perfect way to implement it. I think it is in line the most with how the rest of the tool has been designed. This keeps a rule in the context of a `Model` or `Source`, while keeping things simple and the user is still able to traverse the graph (e.g. from `model_z` it would be possible to retrieve `parent_y`, which in turn has `parent_x`, ...). It also should require no changes to the `rule` interface, just (breaking) changes to the `Model` and `Source` interfaces. 

> Aside: Another half-formed thought here is that this remains very centered around considering a model and its immediate neighbors, rather than the graph as a whole. I think this feels appropriate; there are certainly more "graph shape"-oriented tests I can imagine, but they're rather different from the existing testing modalities in this project and feel reasonable to set aside (for now), I think?

Totally agreed, let's keep it scoped to your request indeed! My feeling is that `graph-shape` oriented-rules are complicated to add and require more thought as everything about `dbt-score` is centered around single entities. Thinking about this, it should probably added in another form (e.g. not using a `rule`). Because with the way `dbt-score` is set up, how will we deduct points from a `Model` or `Source` when the rules was applied to a `Manifest`? Just a thought!

ross-whatnot:
@jochemvandooren thanks for the thoughtful response! I agree that
> This keeps a rule in the context of a Model or Source, while keeping things simple and the user is still able to traverse the graph (e.g. from model_z it would be possible to retrieve parent_y, which in turn has parent_x, ...).
would be the most ergonomic from an end user perspective, but I am somewhat curious how that'll scale. 

For context, I've got a dbt project with ~3000 models in it, and models can have dozens of `1+` upstream model dependencies and hundreds of `+` upstream model dependencies. If models are fully nested (`depends_on` is a list of `Model`s, each of which has their own `depends_on` with a list of `Model`s, etc), things are going to get quite large quickly. That may or may not be a problem in practice, but feels potentially risky at scale. I'm happy to sketch out what that'd look like, but curious if the concern about scale resonates?

Some alternatives:
* Embed `Model`s in `depends_on`, but only `1+` models / strip out the upstream dependencies in those models; this would allow for a lot of rules, but not full graph traversal. Not sure if that's just an awkward compromise, though.
* I also put together [this](https://github.com/PicnicSupermarket/dbt-score/pull/107) as an alternate sketch, which would gather relatives only if / when requested. That might allow a balance wherein _most_ rules (that don't request dependencies) run exactly as they do today, but if rules _do_ want parents / ancestors, they'd have some additional cost associated with doing so.

jochemvandooren:
> @jochemvandooren thanks for the thoughtful response! I agree that
> 
> > This keeps a rule in the context of a Model or Source, while keeping things simple and the user is still able to traverse the graph (e.g. from model_z it would be possible to retrieve parent_y, which in turn has parent_x, ...).
> > would be the most ergonomic from an end user perspective, but I am somewhat curious how that'll scale.
> 
> For context, I've got a dbt project with ~3000 models in it, and models can have dozens of `1+` upstream model dependencies and hundreds of `+` upstream model dependencies. If models are fully nested (`depends_on` is a list of `Model`s, each of which has their own `depends_on` with a list of `Model`s, etc), things are going to get quite large quickly. That may or may not be a problem in practice, but feels potentially risky at scale. I'm happy to sketch out what that'd look like, but curious if the concern about scale resonates?

I do not think the scaling will be too much of an issue actually :thinking: , all models are loaded anyway and we already have the manifest fully loaded anway :thinking: I am afraid there's only one way to find out! 

> * Embed `Model`s in `depends_on`, but only `1+` models / strip out the upstream dependencies in those models; this would allow for a lot of rules, but not full graph traversal. Not sure if that's just an awkward compromise, though.

I think this should be it indeed! My idea would be that a `Model` or `Source` get an extra property. e.g. `parents`. `parents` should be loaded by finding all nodes that are defined in `depends_on`. These are only the first-order parents btw, so I do not think this will become a huge list. I am not sure if we also need `children`, now that I am thinking of it. Also dbt does not have such a property in the manifest for models and sources.

> * I also put together [this](https://github.com/PicnicSupermarket/dbt-score/pull/107) as an alternate sketch, which would gather relatives only if / when requested. That might allow a balance wherein _most_ rules (that don't request dependencies) run exactly as they do today, but if rules _do_ want parents / ancestors, they'd have some additional cost associated with doing so.

Yeah this is interesting indeed, but I am not a big fan of the added complexity of the rules. I do not think this interface is very easy to use, and think the suggestion above is while providing almost the same functionalities

PR #107, opened 2025-03-29
Alternate approach to providing access to parents

ross-whatnot:
As an alternative to my [other PR](https://github.com/PicnicSupermarket/dbt-score/pull/106), this approach feels a bit more ergonomic from an end user perspective; you can define a rule like:
```
@rule
def rule_with_parents(model: Model, parents: Parents) -> RuleViolation | None:
    if 'special' in model.tags and not all(['special' in parent.tags for parent in parents.models]):
        return RuleViolation(message="Special model depends on non-special models")
    return None
```
If the signature includes `parents: Parents`, you will receive a `Parents` object (alias for `Relatives`, which is a dataclass containing `models: set[Model]` and `sources: set[Source]`), so you can write tests about a model in the context of its parents.

This should be easily extensible to other kinds of relatives (`Children`, or with some additional thought about how to represent distance, `Ancestors` or `Descendants`).

Some things I'm not wildly confident around:
* Right now, this calculates the parents only when they're requested (via `get_parents()`); this might not be the best way to do this. My hope is that it'd be easy enough to shift it to being pre-calculated (on creation of the `ManifestLoader`), or to actually build a graph with networkx or something like that that might be more performant; less focused at the moment on the mechanics of calculating the parents than having that get done in the right place.
* The type aliasing around `ModelRuleEvaluationType` and `SourceRuleEvaluationType` feels not quite right; it works, but feels like there's likely a better way to be able to articulate "you might include `parents` (or `children`...)"
* Rather than having `Relatives` as a wrapper around lists of models / sources, could instead just do like `parents: list[Model | Source]`; that does, though, require the rule to figure out what type of parent it's dealing with.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,24 @@
+install:
+	pre-commit install
+	pdm install --group :all
+
+lint:
+	pdm run ruff check .
+	pdm run ruff check --fix
+	pdm run mypy .
+	pdm run tox -e lint
+
+format:
+	pdm run ruff format .
+
+test:
+	pdm run tox -e py
+	pdm run pytest
+	pdm run coverage run -m pytest
+
+docs:
+	pdm run mkdocs build
+	pdm run mkdocs serve
+
+pre-commit:
+	pre-commit run --all-files
--- a/src/dbt_score/__init__.py
+++ b/src/dbt_score/__init__.py
@@ -1,6 +1,6 @@
 """Init dbt_score package."""
 
-from dbt_score.models import Model, Source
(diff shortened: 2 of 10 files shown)


matthieucan:
I think I prefer the approach of a `model` being self-contained and have attributes to access its parents. This can become quite complex to access anything else than direct parents, and I can foresee that not all use-cases can be solved that way (e.g. how to access parents with a distance of both 1 and 2? how to access non-duplicated grand-parents? and other complex concerns which might be better solved in code rather than function signatures)

Issue #118, opened 2025-05-21
Ability to run _only_ selected rules


ross-whatnot:
As we've been stretching our use cases for dbt-score, I've found a few circumstances where I want to be able to run rules in a more targeted way for certain situations. For a concrete example, we have a couple of checks for "are you ref'ing deprecated models", where we want the behavior to be roughly:
* a non-required CI check fails if you have modified files that ref deprecated models
* a required CI check fails if you have added files that ref deprecated models

Currently, I've got a rule for each of these in a dedicated namespace, and execute dbt-score to evaluate that namespace and exclude the _other_ rule (so, the 'modified files' check is run to disable the 'added files' check, and vice versa). But it'd be easier and more flexible in a lot of ways to be able to run dbt-score and have it run _just_ the specific rule(s) I specify at runtime.

My first thought here would be that a behavior could be, roughly:
* add a `--selected-rule` flag that takes a list of rules to run, similar to `--disabled-rule`
* if that's present, rules in the `--selected-rule` are run, ignoring any normal configuration that would disable them
* `--selected-rule` and `--disabled-rule` are mutually exclusive, you can't supply both in a given CLI call

That would allow one to have "special" rules that _aren't_ run during a basic CLI call, but call those rules explicitly and flexibly when needed.

PR #120, opened 2025-06-05
adding `--selected-rule`

ross-whatnot:
As discussed in https://github.com/PicnicSupermarket/dbt-score/issues/118, adding a `--selected-rule` CLI flag to enable only specific rules to be run, either for testing during rules development or in specialized cases for wanting to run narrowly scoped rulesets as a part of CI.

Intended behavior:
* adding `--selected-rule` CLI flag, which is mutually exclusive with `--disabled-rule`
* If `--selected-rule` is provided, the list of selected rules will be executed, even if they are disabled by pyproject.toml

Added some tests, and manual validation confirms it appears to be working as expected.

### DIFF
--- a/src/dbt_score/cli.py
+++ b/src/dbt_score/cli.py
@@ -67,6 +67,12 @@ def cli() -> None:
     multiple=True,
 )
 @click.option(
+    "--selected-rule",
+    help="Rule to select.",
+    default=None,
+    multiple=True,
+)
+@click.option(
     "--manifest",
     "-m",
     help="Manifest filepath.",
@@ -121,6 +127,7 @@ def lint(  # noqa: PLR0912, PLR0913, C901
     select: tuple[str],
     namespace: list[str],
     disabled_rule: list[str],
+    selected_rule: list[str],
     manifest: Path,
     run_dbt_parse: bool,
     fail_project_under: float,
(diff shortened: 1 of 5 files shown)


src/dbt_score/cli.py
```diff
@@ -136,12 +143,19 @@ def lint(  # noqa: PLR0912, PLR0913, C901
     if manifest_provided and run_dbt_parse:
         raise click.UsageError("--run-dbt-parse cannot be used with --manifest.")
 
+    if len(selected_rule) > 0 and len(disabled_rule) > 0:
```
jochemvandooren, line 146: Nice :ok_hand: 

We also need to take into account `pyproject.toml` configuration I think. 

Theoretically I could...

```toml
[tool.dbt-score]
disabled_rules = ["dbt_score.rules.generic.columns_have_description"]
selected_rules = ["dbt_score.rules.generic.model_has_description"]
```

OR

```toml
[tool.dbt-score]
disabled_rules = ["dbt_score.rules.generic.columns_have_description"]
```

`dbt-score lint --selected-rule dbt_score.rules.generic.model_has_description`


Which should not be possible IMO. So maybe we should add more validation to `Config.validate()`?

src/dbt_score/rule_registry.py
```diff
@@ -79,7 +79,13 @@ def _add_rule(self, rule: Type[Rule]) -> None:
         rule_name = rule.source()
         if rule_name in self._rules:
             raise DuplicatedRuleException(rule_name)
-        if rule_name not in self.config.disabled_rules:
+        if (
```
jochemvandooren, line 82: Could you add a small comment that explains this? Our future selves will be grateful :joy:

tests/test_rule_registry.py
```diff
@@ -32,6 +32,33 @@ def test_disabled_rule_registry_discovery():
...
+    config.selected_rules = ["tests.rules.nested.example.rule_test_nested_example"]
+    r = RuleRegistry(config)
+    r._load("tests.rules")
+    assert sorted(r._rules.keys()) == [
+        "tests.rules.nested.example.rule_test_nested_example"
+    ]
+
+
+def test_selected_and_disabled_rule_registry_discovery():
+    """Ensure selected rules are run even if otherwise disabled.
```
jochemvandooren, line 47: Nice I see it does *work* as expected, still I think my comment makes sense; shouldn't we warn the user about this invalid config? It will be less magical to the users I think

jochemvandooren (commented):
Super nice @ross-whatnot ! Thanks again for your contribution :rocket: 

I left a small comment, and two more things:

- Could you add a changelog entry?
- Add some docs to https://dbt-score.picnic.tech/configuration/ ?

PR #122, opened 2025-07-04
Fix column test matching when using quotes in BigQuery

alangner:
I noticed the following issue in BigQuery, when using models where a column has "quote: true" defined.
In that case the data_tests defined for that column were not matched to that node. 
Consequently the test has_uniqueness_test failed.

When looking into the dbt manifest I saw that in the test node at test_metadata.kwargs.column_name the name contained the quotes (with grave accent `), e.g. 
`"column_name": "`my_column`"`
However at the model node it is always without the quotes, e.g.
```
"columns": {
  "my_column": {
    "name": "my_column"
```

### DIFF
--- a/src/dbt_score/models.py
+++ b/src/dbt_score/models.py
@@ -146,7 +146,8 @@ class HasColumnsMixin:
                     for test in test_values
                     if test.get("test_metadata", {})
                     .get("kwargs", {})
-                    .get("column_name")
+                    .get("column_name", "")
+                    .strip("`")
                     == name
                 ],
             )


src/dbt_score/models.py
```diff
@@ -146,7 +146,8 @@ def _get_columns(
                     for test in test_values
                     if test.get("test_metadata", {})
                     .get("kwargs", {})
-                    .get("column_name")
+                    .get("column_name", "")
+                    .strip("`")
```
matthieucan, line 150:
```suggestion
                    .strip("`")  # BigQuery connector when "quote: true"
```
Question: is this a bug in dbt's BigQuery connector? If so, can we include a reference to an issue here?

matthieucan (commented):
Hi @alangner,
Thank you so much for contributing and fixing this bug!
Could you also add an entry in the changelog file?

src/dbt_score/models.py
alangner: I couldn't find a related issue in dbt-adapters, and I am not sure if this is an unwanted behaviour from the connector.

PR #129, opened 2025-08-01
Fix JSON formatter name collision issue

ross-whatnot:
## Summary

- Fix JSON formatter to use `unique_id` instead of `name` as dictionary keys
- Prevents evaluables of different types with the same name from overwriting each other
- Adds comprehensive test case for name collision scenario
- Updates existing tests to match new output format

## Problem

The JSON formatter was using `evaluable.name` as the dictionary key, which caused issues when evaluables of different types (e.g., a model and an exposure) had the same name. The second evaluable processed would overwrite the first in the JSON output.

## Solution

Changed the JSON formatter to use `evaluable.unique_id` (e.g., `model.package.model_name`, `exposure.package.exposure_name`) instead of just the name. This approach:

- Ensures unique keys for all evaluables regardless of name collisions
- Makes the JSON formatter consistent with the Manifest formatter behavior
- Preserves all evaluables in the output

## Changes

- Modified `src/dbt_score/formatters/json_formatter.py` to use `unique_id` as keys
- Updated docstring examples to reflect the new key format
- Updated existing tests in `tests/formatters/test_json_formatter.py`
- Added new test case `test_json_formatter_name_collision_prevention` to verify the fix

## Test plan

- [x] Existing JSON formatter tests pass with updated expected output
- [x] New test case specifically validates name collision prevention
- [x] All linting and type checking passes
- [x] Pre-commit hooks pass

🤖 Generated with [Claude Code](https://claude.ai/code)


jochemvandooren (approved):
Great fix, thanks a lot! :raised_hands:

PR #131, opened 2025-09-18
Improve readme with extended information

matthieucan:


matthieucan:
> Also I think showcasing the output at the start of the README makes a lot of sense; when I look into new tools I always look for a practical use-case first, before reading the actual README

I added another (simpler) example at the top. PTAL!

PR #157, opened 2025-12-19
docs: contributing pre-commits edits for uv

mattcristal:
## What was done

Adjusted `CONTRIBUTING.MD`, the section about pre-commits. Since we are using `uv`, this should be the right way to install and run `pre-commits`?

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -66,7 +66,7 @@ After cloning the repository with git, configure your development environment by
 running these commands from the project's root:
 
 ```shell
-pre-commit install
+uv pip install pre-commit
 uv sync --all-groups
 ```
 
@@ -131,7 +131,7 @@ Cheatsheet:
 Execute hooks manually:
 
 ```shell
-pre-commit run --all-files
+uv run pre-commit run --all-files
 ```
 
 Create a commit bypassing hooks:


docs/contributing.md
```diff
@@ -66,7 +66,7 @@ After cloning the repository with git, configure your development environment by
 running these commands from the project's root:
 
 ```shell
-pre-commit install
+uv pip install pre-commit
```
druzhinin-kirill: I believe there is a little miss interpretation of what this instruction is doing:
- `install pre-commit` - Install pre-commit tool ([this](https://pre-commit.com/#install))
- `pre-commit install` - Install project hooks with pre-commit ([this](https://pre-commit.com/#usage))

As package maintainers, we do not need to be opionated how to install pre-commit system wide (with pip, pipx, uvx, etc.). `pre-commit` is not a dev dependency of this project and optional to use.

Also, we put it as a [recommended prerequisite ](https://github.com/PicnicSupermarket/dbt-score/blob/master/docs/contributing.md#prerequisites), so the further guide assumes it's globally available in your system.

Hope it make sense!

michael-the1:
Closing this PR. I think https://github.com/PicnicSupermarket/dbt-score/pull/157#discussion_r2669052622 gives enough context.

PR #183, opened 2026-06-18
Introduce matrix testing for supported dbt versions

michael-the1:
Add support for supported dbt versions, namely 1.5, 1.6, 1.7, 1.10, and 1.11.

1.8 and 1.9 are omitted because they are [deprecated](https://docs.getdbt.com/docs/dbt-versions?version=2.0&name=Fusion) at the time of writing.


tox.ini
```diff
@@ -8,8 +8,13 @@ runner = uv-venv-lock-runner
 description = Run tests
 dependency_groups = test
 commands =
-   pytest
-   coverage run -m pytest
+    dbt15: uv pip install --upgrade 'dbt-core>=1.5,<1.6'
+    dbt16: uv pip install --upgrade 'dbt-core>=1.6,<1.7'
+    dbt17: uv pip install --upgrade 'dbt-core>=1.7,<1.8'
+    dbt110: uv pip install --upgrade 'dbt-core>=1.10,<1.11'
+    dbt111: uv pip install --upgrade 'dbt-core>=1.11,<1.12'
```
matthieucan: There might be a better way than hardcoding every version, e.g. `uv pip install 'dbt-core~=$dbt_version.0'`, with `$dbt_version = 1.11`. Did you attempt this?

michael-the1: Changed to use environment variables. Looks a bit better now.

matthieucan: It did not work to directly use the value without having to maintain the mapping separately?

michael-the1: I tried a couple things but didn't get very far. You cannot reference the value directly, you something that eventually expresses the form of "dbt-core ~= 1.10.0". So this requires the dots and knowing that you refer to the dbt version.

I tried things with [named factors](https://tox.wiki/en/latest/reference/config.html#factor-label-substitution) but couldn't get it to work.

matthieucan: I see! All good then

tox.ini
```diff
@@ -1,15 +1,21 @@
...
+    dbt15: DBT_VERSION=1.5
+    dbt16: DBT_VERSION=1.6
+    dbt17: DBT_VERSION=1.7
+    dbt110: DBT_VERSION=1.10
+    dbt111: DBT_VERSION=1.11
 commands =
-   pytest
-   coverage run -m pytest
+    dbt{15,16,17,110,111}: uv pip install --upgrade 'dbt-core~={env:DBT_VERSION}.0'
+    coverage run -m pytest
```
druzhinin-kirill: I would consider applying different approach:
- we can have multiple dev dependency groups in uv, also mutually exclusive one defined with [conflicts](https://docs.astral.sh/uv/reference/settings/#conflicts)
- In CI, we can control used dbt version by pointing to the right group in the `dependency_groups` of the `uv-venv-lock-runner`.

This should be less brutal and give us more control over dev environment in general with respect to the lockfile.

michael-the1: So I'm not opposed but just want to note the pros/cons of this approach. You didn't mention it in your message, but I think the approach works best if we also commit a `uv.lock` file.

Pro: reproducible testing environment
Con: we're always testing against a specific version, so we don't automatically test against the latest version. Ex: if we lock against dbt 1.10.0, we have to explicitly update the lockfile to 1.10.1 first and then test again. If we don't use a lockfile, it will happen automagically.

So in this scenario, we'd have to update the lockfile regularly.

Let me know if I don't quite understand this right.

michael-the1: Wait, I didn't realise `uv.lock` was already committed. I guess that means we were already living with this con haha

.github/workflows/ci.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     strategy:
       matrix:
         python: ["3.10", "3.11", "3.12", "3.13"]
+        dbt-core: ["15", "16", "17", "110", "111"]
```
matthieucan, line 22: Not 1.8 and 1.9?

michael-the1: https://docs.getdbt.com/docs/dbt-versions?version=2.0&name=Fusion they're deprecated, should we still support them?

matthieucan: I suppose not, but in this case let's remove 1.5, 1.6 and 1.7. My comment was mostly about the gap between 1.7 and 1.10 in the list

matthieucan: (that said, no harm in supporting one or two deprecated versions, pretty sure they are still used a lot out there)

.github/workflows/ci.yml
```diff
@@ -61,4 +62,21 @@ jobs:
         run: npx --yes prettier@3.1.0 --check "**/*.{json,yaml,yml,md}"
       - name: Run Tox
         run: |
-          uv run tox -e py,lint
+          PYTHON_VERSION=$(echo "${{ matrix.python }}" | sed 's/\.//')
+          uv run tox -e py${PYTHON_VERSION}-dbt${{ matrix.dbt-core }}
+
+  lint:
```
druzhinin-kirill, line 66: Let's move `prettier` step run into `lint`. This will run it only once instead of per matrix entry. Also, it fits `lint` section well 👍

michael-the1: Good idea!

.github/workflows/ci.yml
michael-the1: I think we should match support to what dbt supports. If someone tried to raise a dbt-score issue for a deprecated dbt version, my first suggestion would be for them to upgrade their dbt versions.

matthieucan: Ah wait, I just noticed the difference between "Deprecated" and "End of life". Some deprecated dbt releases are newer than EOL releases, that's very confusing.

matthieucan (approved):

michael-the1:
I updated the merge requirements such that all matrix test builds must pass

PR #185, opened 2026-06-25
fix: Fix Github Actions

michael-the1:
After some org-wide security policy changes, we fix the Github Actions by:
- Pinning the version to a specific SHA 📍
- Changing some third-party actions to homegrown CLI commands 🌱

This PR mirrors the current behavior of our Github Actions. In a future PR, we can consider upgrading the versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,10 +57,7 @@ jobs:
         run: |
           uv sync --locked --only-dev
       - name: Run Prettier
-        uses: creyD/prettier_action@v4.3
-        with:
-          dry: True
-          prettier_options: "--check **/*.{json,yaml,yml,md}"
+        run: npx --yes prettier --check "**/*.{json,yaml,yml,md}"
       - name: Run Tox
         run: |
           uv run tox -e py,lint
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
         with:
           app-id: ${{ secrets.APP_ID }}
           private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          permission-contents: write
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
         with:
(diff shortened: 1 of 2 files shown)


matthieucan (approved):

.github/workflows/ci.yml
```diff
@@ -57,10 +58,7 @@ jobs:
         run: |
           uv sync --locked --only-dev
       - name: Run Prettier
-        uses: creyD/prettier_action@v4.3
-        with:
-          dry: True
-          prettier_options: "--check **/*.{json,yaml,yml,md}"
+        run: npx --yes prettier@3.1.0 --check "**/*.{json,yaml,yml,md}"
```
npeshkov, line 61: Let's fingerprint `prettier`'s version

michael-the1: The SHA fingerprint only applies to referring to other Github actions.

Here we already use an exact pin. If we want an exact lockfile, we'd have to commit a `package-lock.json`.

npeshkov: If it is worth using a fingerprinted `prettier`, I beleive we can install from the package file, bypassing the need for `pyproject.toml` or `package-lock.json`

michael-the1: You mean _just_ using `package.json`? There's no advantage there, the fingerprint comes from `package-lock.json`.

npeshkov: I mean `npx --yes --package=` pointing to git commit for example. I originally thought about the `tar` file but there is no way to fingerprint it ([docs](https://docs.npmjs.com/cli/v11/using-npm/package-spec))

michael-the1: AFAICT, once a package publishes a version on NPM, that is a stable version of the code. So because we pin to a specific version, this will always resolve to the same code. This is different from GHA, where an author can move the version tag to a different commit.

So IMO, no need to point to a specific commit.

`package-lock.json` does help with security here since it does some additional client-side verification.

npeshkov: Alright then!

npeshkov (approved):
👍

Issue #184, opened 2026-06-18
Feature Request: DAG-aware scoring : weight model scores by downstream fan-out


RajSidwadkar:
## What problem does this solve?

Right now, `score_aggregate_evaluables` computes the project score as a flat 
mean across all models. A staging model that feeds 40 downstream nodes and a 
leaf mart model that feeds zero both contribute identically to that number.

In practice, data teams doing triage don't ask "which models have low scores" 
they ask "which violations are causing the most downstream risk." The current 
scoring doesn't help answer that.

## The idea

Since `children` was added to `Model` (and other entities) in #113, the full 
DAG is already available in memory after the manifest is loaded. I'd like to 
propose building on that in a few incremental steps:

**Step 1 : Expose `downstream_count` on `Model`**  
A computed attribute derived from a single BFS pass over the loaded manifest, 
counting direct + transitive descendants. Small, standalone, and immediately 
useful for anyone writing custom rules (e.g. "flag any high fan-out model 
without a contract").

**Step 2 : Opt-in DAG-weighted project scoring**  
A new `scoring_mode = "dag_weighted"` config option that changes 
`score_aggregate_evaluables` to weight each model's score by its 
`downstream_count` instead of a flat mean. Opt-in so existing behaviour is 
untouched.

**Step 3 : `--show critical` CLI flag**  
Sorts and surfaces violations ranked by `downstream_count × severity` gives 
teams an immediate triage list without needing to read through every model.

**Step 4 (optional) :New built-in rule**  
Flag any model whose `downstream_count` exceeds a configurable threshold `N` 
but lacks a dbt `contract` defined. Contracts have been stable since dbt-core 
1.5, which is already dbt-score's minimum requirement.

## Why not just a custom rule?

Steps 2 and 3 require changes to core scoring/CLI behaviour, which can't be 
done from a custom rule. Step 1 could theoretically be replicated in a custom 
rule today by traversing `children` manually, but having it as a first-class 
attribute feels like the right abstraction given that `parents` and `children` 
are already there.

## Proposed starting point

Happy to open a small PR for just Step 1 (adding `downstream_count` to `Model`) 
if the direction makes sense to you that way we can validate the approach 
before anything touches scoring or CLI.

Would love to hear if this has been considered before or if there are concerns 
about scope.

matthieucan:
Hi @RajSidwadkar, thank you for writing down this proposal!

> Step 1 : Expose downstream_count on Model

[...]

PR #188, opened 2026-07-10
Add downstream_count attribute via BFS DAG traversal

RajSidwadkar:
## Summary

Adds a `downstream_count` property to `Model`, computed via BFS over the DAG.

This is Step 1 of the proposal in #184.

## Context

From the discussion in #184, I said:

> Step 1: Expose downstream_count on Model
> A computed attribute derived from a single BFS pass over the loaded manifest, counting direct + transitive descendants. Small, standalone, and immediately useful for anyone writing custom rules (e.g., "flag any high fan-out model without a contract").

This PR is that follow-up implemented, tested against synthetic DAG structures, and validated against a real dbt manifest.

## Changes

- `src/dbt_score/models.py`: new `downstream_count` property on `Model`, BFS traversal over `children`, deduped via a visited set (handles diamond-shaped DAGs without double-counting).
- `tests/test_models.py`: 5 new test cases (isolated model, single child, linear chain, diamond DAG, non-Model descendants).

## Testing

- `uv run pytest` - 129 passed
- `uv run mypy .` - clean on touched files (1 pre-existing, unrelated error in `dbt_utils.py`)
- `uv run ruff check .` - clean

Validated against a real dbt manifest (`dbt-labs/jaffle_shop_duckdb`) and cross-checked independently against `dbt ls --select <model>+`:

| Model | downstream_count | dbt ls --select <model>+ (minus itself) |
|---|---|---|
| customers | 0 | 0 |
| orders | 0 | 0 |
| stg_customers | 1 | 1 |
| stg_orders | 2 | 2 |
| stg_payments | 2 | 2 |

All values match independently, confirming the BFS traversal and dedup logic behave correctly against a real manifest, not just synthetic test fixtures.

## Notes

Part of #184 - Step 4 (high-fan-out contract rule) will follow as a separate PR once this lands, since it depends on this attribute.

### DIFF
--- a/src/dbt_score/models.py
+++ b/src/dbt_score/models.py
@@ -3,7 +3,7 @@
 import json
 import logging
 import re
-from collections import defaultdict
+from collections import defaultdict, deque
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Iterable, Literal, TypeAlias, Union
@@ -264,6 +264,35 @@ class Model(HasColumnsMixin):
     def __hash__(self) -> int:
         """Compute a unique hash for a model."""
         return hash(self.unique_id)
+
+    @property
+    def downstream_count(self) -> int:
+        """Calculate the count of all downstream model dependents.
+
+        Returns:
+            int: The count of unique downstream models.
+        """
+        visited: set[str] = set()
+        queue: deque[ChildType] = deque()
+        for child in self.children:
+            if child.unique_id not in visited:
+                visited.add(child.unique_id)
+                queue.append(child)
+
+        count = 0
+        while queue:
+            current = queue.popleft()
+
+            if isinstance(current, Model):
+                count += 1
+
+            if hasattr(current, "children"):
+                for grandchild in current.children:
+                    if grandchild.unique_id not in visited:
+                        visited.add(grandchild.unique_id)
+                        queue.append(grandchild)
+
+        return count
 
 
 @dataclass
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -3,7 +3,7 @@
 from pathlib import Path
 from unittest.mock import patch
 
-from dbt_score.models import ManifestLoader
+from dbt_score.models import Exposure, ManifestLoader, Model, Snapshot
 
 
 @patch("dbt_score.models.Path.read_text")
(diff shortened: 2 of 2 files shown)


src/dbt_score/models.py
```diff
@@ -265,6 +265,34 @@ def __hash__(self) -> int:
...
+
+            if isinstance(current, Model):
+                count += 1
+
+            if hasattr(current, "children"):
+                for grandchild in current.children:
+                    if grandchild.unique_id not in visited:
+                        visited.add(grandchild.unique_id)
+                        queue.append(grandchild)
+
```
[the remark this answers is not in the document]
RajSidwadkar: Applied - switched to deque.popleft() and unified the initial/BFS enqueue logic behind a single visited check (fixes the duplicate-children edge case too). Ran a benchmark on synthetic linear DAGs up to 16k nodes: the old list.pop(0) version degrades quadratically, the new version stays linear (~Nx speedup at N=16000, see attached script). For typical dbt project sizes (hundreds to a few thousand models) this is comfortably a single fast BFS pass.

RajSidwadkar:
Following up on the compute-time question - I ran two benchmarks to check this properly.

**Realistic case:** a layered DAG (5 layers x 400 models, branching dependencies - the shape a real dbt project has), 2000 models total. Computing downstream_count for every single model in the manifest (worst-case usage: a rule that checks it on every model) took ~9-11ms total, against a manifest load time of ~47-67ms for comparison. That's roughly 17-22% on top of a cost already paid just to load the manifest, and per-call average works out to ~0.005ms.

**Worst case (for full disclosure):** I also tested a single 2000-model linear chain, which is the pathological case for this approach, since computing downstream_count independently for every node in a chain of depth N is O(N^2) overall. That case took ~1070ms for the same "every model" scenario. This shape doesn't reflect real dbt projects (which are shallow and layered, not deep linear chains), but wanted to flag it rather than only show the flattering number.

Happy to add a note in the docstring calling out this Big-O characteristic if that's useful for future maintainers, or explore memoization if you'd rather guard against the adversarial case too - let me know which you'd prefer.

.gitignore
michael-the1, line 1: Can you remove this change? It looks specific to your setup.

To ignore a file only for you, you can modify the `.git/info/exclude` file.

michael-the1 (commented):
LGTM! I only have a minor nit on the `.gitignore`. Thank you for your contribution!

Thanks especially for the performance benchmark you ran. I don't think memoization is necessary. We can consider it for the future.

.gitignore
RajSidwadkar: Thanks for the review! 

I've removed the local directory from .gitignore as requested and moved it to my local .git/info/exclude file instead. 

The PR is updated with a clean commit history!

michael-the1 (approved):
Great work, thank you for your contribution!

PR #193, opened 2026-07-15
fix: honor a threshold of 0 for --fail-project-under / --fail-any-item-under

michael-the1:
## What changed

`dbt-score lint` silently ignored `--fail-project-under 0` and
`--fail-any-item-under 0`. The CLI used a truthiness check
(`if fail_project_under:`) to decide whether to apply the override, and since `0`
is falsy the value was dropped and the default (`5.0`) was used instead.

Changed the checks to `is not None`, so an explicit `0` (a valid threshold) is
respected. Added a regression test.

Files: `src/dbt_score/cli.py`, `tests/test_cli.py`.

## How to review

- Confirm the `is not None` logic is correct for all option combinations
  (default unset vs. `0` vs. positive values).
- Run the regression test:
  ```shell
  uv run pytest tests/test_cli.py -k fail_thresholds_zero
  ```
- Manual sanity check: `dbt-score lint --fail-project-under 0 ...` on a project
  should no longer fail purely due to the default 5.0 threshold.
- Human input: confirm the intended semantics — `0` should be a valid, honored
  threshold.

### DIFF
--- a/src/dbt_score/cli.py
+++ b/src/dbt_score/cli.py
@@ -130,8 +130,8 @@ def lint(  # noqa: PLR0913, C901
     disabled_rule: list[str],
     manifest: Path,
     run_dbt_parse: bool,
-    fail_project_under: float,
-    fail_any_item_under: float,
+    fail_project_under: float | None,
+    fail_any_item_under: float | None,
     show: Literal["all", "failing-items", "failing-rules"],
     debug: bool,
 ) -> None:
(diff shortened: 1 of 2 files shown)


src/dbt_score/cli.py
```diff
@@ -149,9 +149,9 @@ def lint(  # noqa: PLR0913, C901
         config.overload({"rule_namespaces": namespace})
     if disabled_rule:
         config.overload({"disabled_rules": disabled_rule})
-    if fail_project_under:
```
matthieucan, line 152: Shouldn't the type above `fail_project_under: float` be updated to take `None` into account?

michael-the1: Yes, good point. Added it.

matthieucan (approved):

PR #195, opened 2026-07-15
test: update test matrix to only combine supported python and dbt versions

michael-the1:
Not all dbt version x python versions are [officially supported](https://docs.getdbt.com/faqs/Core/install-python-compatibility). While all tests still used to pass, in some tests in https://github.com/PicnicSupermarket/dbt-score/pull/194 these unsupported combinations started failing. Therefore, update the matrix to only combine supported versions.

I've decided to duplicate the matrix in `ci.yml` and `tox.ini`. We could do some magic to have only the env list in `tox.ini`, but it was a bit too much magic for my tastes. Instead: add a comment to `ci.yml` indicating this duplication.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,9 +17,24 @@ jobs:
   build:
     runs-on: ubuntu-slim
     strategy:
+      fail-fast: false
+      # NOTE: This matrix duplicates the supported Python/dbt-core combinations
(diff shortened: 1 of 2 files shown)


jochemvandooren (approved):
👌

michael-the1:
I updated the required tests in the Github PR checks to reflect this change.

PR #206, opened 2026-07-16
Sign uv-bot commits

sercancicek:
Swaps the `git commit`/`git push` step for a `gh api` Contents API call, so the Dependabot lockfile-refresh commit shows as Verified.


michael-the1 (approved):
LGTM!

PR #215, opened 2026-07-24
feat: make severity values fully configurable across all four levels

### LINKED ISSUE #209: Feature Request: Configure score weight
Currently, the scoring weights are linear: low = 1, medium = 2, high = 3, and critical = 4.

With more than 20 rules in a project, the score difference between one high and one low rule becomes too small. For example, if there is a single breaking rule marked high, the overall model score may be 9.8, while the same issue marked low may result in 9.9. In practice, this makes it possible to “game” the system by fixing all lower-priority rules and still passing the score threshold.

However, in real-world use, a high-priority rule may be far more important and require much more attention. The request is to support configurable rule weights so teams can define their own scoring priorities. For example, one team might use low = 1, medium = 10, high = 100, and critical = 1000. This would allow different teams to tune the scoring system to match their priorities.

pikachuev:
Users can now define a unified numeric scale for all severity levels — `low`, `medium`, `high`, and `critical` — under `[tool.dbt-score.severity_values]`. Previously only three levels were configurable and their numeric scale was separate from how severity was assigned to individual rules.

The rule config option `severity` now uses the same number space as the configured values, so the scale is consistent across the whole configuration file. For example, setting `critical = 10` means rules can be assigned `severity = 10` to mark them as critical.

The `critical` severity retains its special behaviour: any violation always forces the score to `0` regardless of its configured value. Values must be strictly increasing (`low < medium < high < critical`).

The section is renamed from `severity_weights` to `severity_values` to better reflect its purpose.

Closes #209


michael-the1:
@pikachuev Thank you for your contribution. FYI I already started looking at it, hoping to post my review next week.

docs/configuration.md
```diff
@@ -75,6 +81,29 @@ All badges except `wip` can be configured with the following option:
...
+- `critical` (default: `4`)
+
+Values must be positive integers and strictly increasing
+(`low < medium < high < critical`). A rule is worth the highest configured value
+in points; a violation costs its severity's value. Note that `critical` is
+special: regardless of its configured value, any critical violation always
+forces the score to `0`.
+
+The default values can be found in the
+[SeverityValueConfig](reference/config.md#dbt_score.config.SeverityValueConfig).
```
matthieucan, line 105: Defaults are already written above (lines 93-96), I think we can skip this link to keep it simple

src/dbt_score/config.py
```diff
@@ -47,6 +47,39 @@ def validate(self) -> None:
...
+        """Validate the severity value configuration."""
+        for name in ("low", "medium", "high", "critical"):
+            value = getattr(self, name)
+            if isinstance(value, bool) or not isinstance(value, int):
+                raise ValueError(f"Severity value {name!r} must be an integer.")
+            if value <= 0:
+                raise ValueError(f"Severity value {name!r} must be greater than 0.")
+        if not (self.low < self.medium < self.high < self.critical):
+            raise ValueError(
+                "Severity values must be strictly increasing:"
```
matthieucan, line 78: Do we need them to be _strictly_ increasing?

src/dbt_score/config.py
```diff
@@ -47,6 +47,39 @@ def validate(self) -> None:
...
 
 
+@dataclass
+class SeverityValueConfig:
+    """Configuration for the value of each rule severity."""
+
+    low: int = 1
+    medium: int = 2
+    high: int = 3
+    critical: int = 4
```
matthieucan, line 57: I appreciate backwards-compatible defaults, but the discussion is worth having: should we take this opportunity to change the defaults to better match reality and prevent the rule-gaming described in #209?

matthieucan (commented):
Thank you for your contribution @pikachuev!

src/dbt_score/config.py
pikachuev: I don’t think there’s a single “right” set of severity values that works for every use case. Some teams might prefer a simple T‑shirt–style scale with roughly doubling steps, like low‑1, medium‑2, high‑8, critical‑16, or low‑1, medium‑5, high‑10, critical‑20. Others might want a more exponential spread, such as low‑1, medium‑10, high‑100, critical‑1000. Which approach makes sense really depends on the specific context and how each team wants to represent the gap in impact between severity levels.

On top of that, many teams already rely on their current severity mappings checks, and introducing new values or changing the scale could break those existing rules.

src/dbt_score/config.py
pikachuev: They only need to be strictly increasing if we’re treating each severity level as inherently more important than the one before it. I did think about allowing two severities to share the same weight, , there is no direct impact, it could just make things confusing.

src/dbt_score/scoring.py
```diff
@@ -29,20 +29,23 @@ def rounded_value(self) -> float:
...
 
+        # Values come from severity. A rule is worth `score_cardinality`
+        # points; a violation costs the severity's value. With the default
+        # values (low=1, medium=2, high=3) and cardinality of 3, a violation
+        # yields a score of 2/3 (low), 1/3 (medium) or 0/3 (high). CRITICAL is
+        # excluded from the cardinality and handled separately: any critical
+        # violation forces the score to 0.
+        self._values = config.severity_value_config.as_dict()
+        non_critical = {k: v for k, v in self._values.items() if k != Severity.CRITICAL}
+        self.score_cardinality = max(non_critical.values())
```
michael-the1, line 47: I also appreciate the backwards-compatible nature of this implementation, but it does make me consider: the value of critical actually doesn't make an impact at all.

So is it worth configuring? I think that just makes it confusing to users.

michael-the1 (commented):
Great contribution!

PR #225, opened 2026-08-30
feat: add group owner to Model (#121)

### LINKED ISSUE #121: feature request: Extract group owner as a Model field
The Model class pulls out the group name, but does not pull out the group owner, if any. I would like to use this in a check for whether the model has an owner or not.

imamAtif:
Closes #121

Model had a `group` string field but no way to check group ownership. Rules that need to verify if a model has an owner or who owns it had to work around this.

## Solution

Added `Group` and `Owner` dataclasses and wired them into `ManifestLoader` so `Model.group_owner` is populated from the manifest `groups` dictionary.

Changes:
* `src/dbt_score/models.py` - Added `Owner` and `Group` with `from_raw_values`, added `Model.group_owner: Owner | None` (kept `Model.group: str | None` for backward compat), added `ManifestLoader.groups` and `_load_groups`, and set `model.group_owner` in `_load_models` via lookup by `group.<package>.<name>`
* `src/dbt_score/__init__.py` - Export `Group` and `Owner`
* `tests/resources/manifest.json` - Added sample groups `them_over_there` with owner Alice and `default` with no owner, plus `group_map`
* `tests/test_models.py` - Added 4 tests for loaded owner, list email, missing group key, and manifest without groups

Owner email keeps the manifest type `str | list[str] | None` so both single and list cases work.

## How to verify

```bash
uv run pytest tests/test_models.py::test_group_owner_loaded -v
uv run pytest tests/test_models.py::test_group_owner_with_list_email -v
uv run pytest tests/test_models.py::test_model_without_group_key -v
uv run pytest tests/test_models.py::test_manifest_without_groups -v

# Full suite
uv run pytest
uv run coverage run -m pytest && uv run coverage report
uv run ruff check src/dbt_score/models.py src/dbt_score/__init__.py
uv run mypy src/dbt_score/models.py
```

Example rule:

```python
from dbt_score import Model, Rule, RuleViolation, rule

@rule
def model_must_have_group_owner(model: Model):
    """Check that the model has a group owner."""
    if model.group and model.group_owner is None:
        return RuleViolation(f"{model.name} has group {model.group} but no owner")
    if model.group_owner and model.group_owner.name is None:
        return RuleViolation(f"{model.name} group owner missing name")
```

## Results

```
135 passed
Coverage 93%
Ruff format clean
Mypy clean on changed files
```


src/dbt_score/models.py
```diff
@@ -219,6 +288,8 @@ class Model(HasColumnsMixin):
     children: list[ChildType] = field(default_factory=list)
     _raw_values: dict[str, Any] = field(default_factory=dict)
     _raw_test_values: list[dict[str, Any]] = field(default_factory=list)
+    group: str | None = None
+    group_owner: Owner | None = None
```
jochemvandooren: I am not sure about adding `group_owner` here, considering only `group` can be configured in models. If we really want to lint `group_owner` I think it makes more sense to add `groups` as a different entity and lint the group metadata as well?

jochemvandooren (commented):
Thanks for opening the PR! 🙌 

I think the code looks good! I do want to question the approach a little bit, if the goal is to lint metadata of groups, I think we should actually add it as another entity in dbt-score, if that makes sense?

Now we are actually mingling `model` and `group` attributes into `model` (i.e. `group_owner`) and I don't think that will be very futureproof

PR #232, opened 2026-09-15
fix: load manifests that omit optional keys

### LINKED ISSUE #231: Fusion (dbt v2) manifests omit optional keys that from_node indexes directly
dbt v2 (the Fusion engine, `dbt==2.0.1`) writes a v12 manifest but leaves out optional keys when they're null, where dbt-core wrote them as explicit `null`s. dbt-score reads six of those with `node_values["..."]` and crashes on the first:

```
  File ".../dbt_score/models.py", line 243, in from_node
    group=node_values["group"],
KeyError: 'group'
```

All six are optional in the v12 schema, each `anyOf: [string, null]` or defaulting to null and none listed under `required`:

- `Model.from_node`: `group`, `patch_path`
- `Source.from_node`: `source_meta`, `patch_path`
- `Snapshot.from_node`: `patch_path`
- `Seed.from_node`: `patch_path`

On a 1097-model project Fusion omits `group` on every model, `patch_path` on the 11 models with no YAML patch and on every source, and `source_meta` on every source. Setting those three keys to `null` in the manifest was enough for `dbt-score lint` to load everything and score normally, so `.get()` on the six reads is the whole fix. I'll open a PR.

dbt-score 0.17.0, dbt 2.0.1 on Snowflake, Python 3.14.

dobsontom:
This PR stops `dbt-score lint` failing with `KeyError: 'group'` on the first model of a manifest written by dbt v2 (Fusion). dbt-core always wrote those keys, as an explicit null where it had no value, so indexing them was safe. Fusion leaves them out.

`Macro.from_node` already indexes only its required keys and reads the rest with `.get()`. This applies the same to `Model`, `Source`, `Snapshot`, `Seed`, `Exposure`, and `Column`, and widens eight annotations to match. Three of those were wrong before Fusion existed: `label`, `url`, and `maturity` are null on all three exposures in `tests/resources/manifest.json` whilst typed `str`. Nothing in `src/` or `docs/` reads any of the eight.

The new test strips the keys a real dbt 2.0.1 project omits (measured across 1097 models) and fails on master.

Two things I have left for you:

- `manifest_formatter.py:37` indexes `meta` on the raw manifest, which is also optional in v12, so a manifest without it loads here and then fails at output. Happy to follow up separately.
- #225 makes the same `group` change. I kept it because it is the key in #231's traceback, so dropping it would leave this not fixing its own issue. I will rebase if #225 lands first.

Closes #231


jochemvandooren (approved):
Very well, thanks for opening the PR! 🙌 Code looks good, tests run well.

I think we can leave `meta` as it is now. Even when omitting `meta` from models, it still appears in the `manifest.json` so no need to fix it (yet)

jochemvandooren:
Just needs to resolve conflicts for the CHANGELOG.md, rest looks good ✨

